PR #27, opened 2018-06-04
Fix variable typo in init script

bskaggs:
Fix variable used to determine if DB is found: GALLERY__MYSQL__USERNAME instead of GALLERY__MYSQL__USER


jeffyjefflabs:
Thanks!

PR #42, opened 2019-02-17
Fix email validation against the allowed_domains list

HansK-p:
Email validation against the allowed_domains list didn't work when there was more than one allowed domain.

This is my first pull request ever, so bear with me if I've made mistakes.


HansK-p:
That was actually an embarrassing (and bad) mistake. Fixed.

jeffyjefflabs:
Thanks for the contribution - sorry for the wait, I somehow hadn't noticed the PR.

PR #71, opened 2019-11-14
Accessibility-6: All decorative elements hidden to screen readers, all metrics and elements without text now have textual descriptions for users with screen readers, and all of site is accessible for those who render site without stylesheet

manfromjupyter:
- All of site is accessible to users that render site without a stylesheet.
- All decorative elements are hidden to screen readers
- All metrics and elements have textual descriptions for users with screen readers
- Code Cell page fix (numbering was off in tables at the bottom of the page)
- Fixed a couple modals that had the submit button outside the form element
- Refactoring editing experience of titles, tags, and descriptions of notebooks.
- Reduced tech debt by replacing duplicated code in a number of places with a new slim file that upon being updated will update both (prevent mismatches).
    - Changed "notebooks home" listings and "notebooks" listings to both inherit the similar code from now "notebook_listings".
    - Changed notebook listing's "nb_stats" and notebook jumbotron's "tab_edit/tag_view" pages that mention metric and stats for the notebooks to now inherit from a single source notebook_metadata/notebook_metadata_metrics (a little hard to explain)
- Fixed order of header elements
- Hid dropdown elements from HTML until they are toggled/opened (reduces number of tabs to get past it in some places and in others, creates less noise to users who view the web app with the stylesheet disabled.


app/views/layouts/layout.slim
```diff
@@ -173,127 +235,144 @@ html lang="en"
...
-          ==render partial: 'modals/login_modal'
-          ==render partial: 'modals/upload.slim'
-          ==render partial: 'modals/environment_loading'
-          ==render partial: 'modals/new_group.slim'
+          -url_check = request.path.split("/")
+          ==render partial: "modals/login_modal"
+          ==render partial: "modals/upload.slim"
+          ==render partial: "modals/environment_loading"
+          ==render partial: "modals/new_group.slim"
           -if url_check[1] == "groups" && url_check[2] != nil && url_check[2][0..3] =~ /\d/
```
mcrutch: Need to check if the group object exists before rendering this.  There are somtimes errors getting thrown here for null objects not having an ID

mcrutch:
fixes #73

PR #84, opened 2019-12-10
User Link Tech Debt fix/cleanup and hover fix when user has no org

manfromjupyter:
- Changed all user links to use the link_to_user helper and included logic for displaying org on title if they exist for said user (so removed a bit of tech debt too)
- Changed the notice flash messages to success flash messages when doing user model changes.
- Is too wordy to include "belongs to org: Engineering" for all user links for screenreader users so decided to just add it to the header on the User Summary page (so a minor change there too).
- Did some catch up for the dark and ultra dark themes

Tested everything but confirming the success message appears after a user is created and their profile is updated, but it does work every time for updating users and creating users and it was of the same code format.

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -2023,11 +2023,18 @@ body.page-metrics {
     }
 }
 #notebookDisplay {
-    pre.highlight {
+    a.cell-health + span.sr-only + span.sr-only + hr.hidden + div pre.highlight {
         margin-top: -1.2em;
     }
     pre.highlight.highlight-text {
         margin-top: 0;
+    }
+    pre.highlight ~ img {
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/styles.scss
```diff
@@ -2702,7 +2776,7 @@ body.dark-theme {
     #topNavbar .logo-link .logo-nb {
         filter: invert(100%);
     }
-    .language-icon {
+    .language-icon[href*="/languages/python"] {
```
mcrutch: Why only for languages/python?

mcrutch: For the record, the reason is so that the Python logo's # (2/3) is visible in display modes with dark backgrounds.

PR #91, opened 2019-12-19
Accessibility-9: Made it so all header elements and dropdowns across the product can now be opened with either 'enter' or 'space' and made it so canceling the editor for the notebook title, tags, or description does not force the page to reload

manfromjupyter:
Accessibility-9: Made it so all header elements and dropdowns across the product can now be opened with either 'enter' or 'space' and made it so canceling the editor for the notebook title, tags, or description does not force the page to reload. In addition added type of search to all search bars because although we probably won't see much happen, it is listed as a potential accessibility improvement and users could maybe do internal searches of a site from their browser url address bar after tabbing one time (if everything is set up right?)

I have tested and it all works the same as always for search and have made sure the rest work given these new requirements.

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -11,7 +11,8 @@ document.addEventListener("turbolinks:load", function(){
   /* ===================================== */
   $('#mobileNavContainer').keypress(function(e){
     var keycode = (e.keyCode ? event.keyCode : event.which);
-    if(keycode == '13'){
+    if(keycode == '13' || keycode == '32'){
+      e.preventDefault();
       $('#mobileNavContainer').click();
     }
   });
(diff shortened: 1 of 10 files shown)


mcrutch:
Can't close Notebooks dropdown with Spacebar.  Tab to the down arrow, hit space bar, opens, hit space bar again, moves down to next element.  If space is going to work for that element it should work both directions.

manfromjupyter:
@mcrutch, I fixed that issue. Apologies. Spaces now can both open and close all of the header dialogs, as well as the Change Requests View dropdown toggle and the Recommendations toggle. I had it working for all but forgot to test it for that notebook dropdown button.

- In addition, because it's still Accessibility related, I added in the role=presentation change for the notebook listings table, changed the font-size of the admin-only snippetInfo, and fixed a typo-o or two.

I have tested it all and even retested all of the previous-tested changes still in this ticket.

PR #104, opened 2020-01-08
Theme fixes 1

manfromjupyter:
- Cleanup to primarily Dark and Ultra Dark Themes
    - Replaced loading images
    - Fixed the issue where there the search icon on home page just looked like a giant black rectangle.
    - Other design enhancements
- Fixed the issue where the success message after saving a new theme would immediately show up then disappear
- Added functionality (regardless of theme) to make the background color change just slightly for select dropdowns when hovered over them and made it so the cursor changes to a pointer, so users can easily differentiate it from an input field.


mcrutch:
Go to edit preferences, change settings, leave flash message up, go to home page, go back to user-preferences.  Flash message briefly appears  still and then vanishes.

PR #107, opened 2020-01-10
Grants notebook editors ability to change ownership to other users

manfromjupyter:
- Refactored the Change Owner notebook action form and modal
- Enabled users to set new owner of a given notebook they have editor rights to to any other user
- Made the form autocomplete for both groups and users based on input
- Added metadata of who the owner is on every notebook jumbotron page (didnt exist before, only showed for groups that owned the notebook - doesn't take into account notebooks where the last editor and owner are two different people)
- Made alerts created by the devise file have some sort of styles (alert-alerts now look like alert-infos (the light blue ones))


mcrutch:
type quickly and press enter - 404 error is presented in modal
type quickly and click change owner before drop-down list appears - changes to current owner

PR #117, opened 2020-01-30
Brought back functionality for admins on private pages and minor tweaks to private icon on jumbotron page

manfromjupyter:
Brought back functionality for admins on private pages and made lock icon on jumbotron pages bigger and moved them near title for product consistency

P.S. Tested and I recommend using notebook # 24 for testing, but really any notebook not yours can be made private then public again no problem so any really work.

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -1558,9 +1558,6 @@ p.description {
     margin-bottom: .85em;
     white-space: pre-line;
 }
-.admin_hidden {
-    opacity: .6;
-}
 .action-icon.view-summary {
     margin-top: 20px;
 }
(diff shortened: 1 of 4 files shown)


mcrutch:
As an Admin, Unless I have permission to the notebook I still shouldn't see comment, feedback, star, subscribe, etc... Just really need the "gear" as admins on a private notebook. Current test notebook is also viewable by me so just loop back and double check that with some more test data before I move forward with this merge.

manfromjupyter:
Made said changes. Note that the majority of the huge influx of changes now are just due to tabbing all of the elements below a couple new if statements, nothing crazy has been done to those blocks of green code/red code in the Jumbotron.

PR #126, opened 2020-02-04
Browser-title-fixes: If objects are returned as nil due to the browser not being modern, made it so it won't error out

manfromjupyter:
Browser-title-fixes: If objects are returned as nil due to the browser not being modern, made it so it won't error out

Majority of changes below are just due to being tabbed due to nesting the code in an if/else.

Tested IE and Chrome.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -184,7 +184,7 @@ class ApplicationController < ActionController::Base
   def setup_browser_titles
     browser = Browser.new(request.env["HTTP_USER_AGENT"])
     if !browser.modern?
-      title = "NBGallery"
+      title = "#{GalleryConfig.site.name} - Error Unsupported Browser"
     else
       url_check = request.path.split("/")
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -182,118 +182,123 @@ def setup_body_classes
...
+        elsif url_check[2] == "recently_executed"
+          title = "Notebooks Recently Executed"
+        elsif url_check[2] == "shared_with_me"
+          title = "Notebooks Shared with Me"
+        else
+          title = "#{url}"
+        end
+      elsif url_check[1] == "groups"
+        if url_check[2] == nil
+          title = "All Groups"
```
mcrutch: What if the group doesn't exist?  Can you still get here?  What about
if(@group != nil)
   title= Group group.name
elsif url_check[2]=nil
  All Groups
else
  title="Groups"
end if

manfromjupyter: I tried to hit all of the endpoints for notebooks, users, groups, and change requests but for every error, the only time the element is not attached is when the browser is not modern.

PR #133, opened 2020-02-10
Search UIUX

manfromjupyter:
- Customized search page with a clean way by parsing out filters and then provides easy ways to add and remove filters.
- Add search tooltips when selecting new info button on non-search page search containers.
- Other minor fixes


manfromjupyter:
It's so close, just need to have to still support adding multiple word searches, touch up home page search bar length, and make info button a toggleable action

manfromjupyter:
@mcrutch, should be good to go now. Added alerts too. Anyway, lemme know

PR #134, opened 2020-02-13
Allow searching by packages found in notebooks. Also made a change in…

mcrutch:
… support of upgrading to Solr 8 (Solr 8 does not allow negative boosting)


mcrutch:
This adds a new approved search field (package)

PR #135, opened 2020-02-13
Support exclusion of packages

mcrutch:


manfromjupyter:
I'm not sure how to test this, but it didn't break current search, so accepted! Hahah

PR #136, opened 2020-02-21
Catchup-4: Catchup

manfromjupyter:
-Fixed an alert issue where it was appearing twice on notebook listing pages
-Fixed a coloring issue for ultra dark theme
-Fixed an z-alignment issue for ultra dark theme
-Cleaned up dark theme a tid bit for the search container
-Fix for snippetInfo for certain types of returns (weird irregular issue)
-Got recommended dropdown elements loading properly again.

### DIFF
--- a/app/models/notebook.rb
+++ b/app/models/notebook.rb
@@ -474,7 +474,7 @@ class Notebook < ActiveRecord::Base
     recommendations = recommendation_snippet
     snippet =
       if highlights && recommendations
-        "#{highlights}<br><br>#{recommendations}"
+        "#{highlights}#{recommendations}"
       elsif highlights
         highlights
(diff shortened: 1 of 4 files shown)


app/models/notebook.rb
```diff
@@ -474,7 +474,7 @@ def snippet(user)
     recommendations = recommendation_snippet
     snippet =
       if highlights && recommendations
-        "#{highlights}<br><br>#{recommendations}"
```
mcrutch: Rather than just remove the <br>, find examples that actually fill out those two and make sure it still displays correct.  Or rather than just saying if x&&y check if they are also non-zero length as well.  This feels like a lazy hack that may cause other display issues.

PR #140, opened 2020-02-29
Deprecate Notebooks Functionality (Admins only so far)

manfromjupyter:
### **CHANGES:**
- Allowed Admins to access a modal to deprecate a notebook and enabled the owner and admins to un-deprecate
 - Added autocomplete functionality for notebook selection
- Removed Tech Debt with alert container
- Prepopulated edit form when editing
- Allowed multiple alternate notebooks to be added
- Removed all download and run in jupyter buttons across the product for notebooks that have been deprecated and opted in to have frozen.
- Added deprecation tag to notebook listings to explain when a notebook is deprecated before they click on it.
deprecated tag

### **BLOCKERS:**
- Did not decrease search ranking for them yet because I was honestly afraid I was going to mess up or do it wrong anyway with all of the recent search changes done. Need to consult @mcrutch what is the best way to do this.
- Did not prevent deprecated notebooks from getting imported to a user's jupyter instance from the starred folder because I need to know if we can edit the /public folder as it seems to be the only way to do it. If so, looks like an ajax call is the best way to get the job just need to know if we can touch that or if it'd just get overwritten.


app/models/notebook.rb
```diff
@@ -20,6 +20,7 @@ class Notebook < ActiveRecord::Base
   has_many :revisions, dependent: :destroy
   has_many :reviews, dependent: :destroy
   has_many :subscriptions, as: :sub, dependent: :destroy
+  has_one :deprecated_notebbook
```
mcrutch: To add the ability to search boost on this, in this block - https://github.com/nbgallery/nbgallery/blob/5ed6b7864c9349ef57e4b35a6b98b34ab441f3e4/app/models/notebook.rb#L35 down near packages, add another field called active or something that is a boolean and set to true if the notebook is NOT active (Easier to positive boost a good value since you can't negatively boost a bad value).  Then we need to toy with the weight further down in the file but we at least need it set.  After this is added, a Notebook.reindex will add it to solr

mcrutch:
For the top of the page.
"This notebook has been deprecated and disabled."
"This notebook has been deprecated."

PR #143, opened 2020-03-02
Remove turbolinks

mcrutch:


manfromjupyter:
I made some edits to allow our modals found in forms to successfully reload so that the success appear when applicable. @mcrutch assigning back to you because I made a hack at the owner modal. I still cannot figure out why it wants to Patch the notebook as a whole using the update method after owner is patched, but this ignores 404 errors (including the one caused by looking up the staged notebook which never exists). Anyway, let me know if you are in favor or not of this hack.

app/controllers/notebooks_controller.rb
```diff
@@ -359,12 +359,15 @@ def owner=
...
+      else
+        user = User.find_by!(user_name: params[:owner])
+        flash[:success] = "Owner of notebook has been set to #{user.first_name} #{user.last_name} successfully."
+      end
     else
-      user = User.find_by!(user_name: params[:owner])
-      flash[:success] = "Owner of notebook has been set to #{user.first_name} #{user.last_name} successfully."
+      render json: @notebook.errors, status: :unprocessable_entity
     end
     redirect_to(:back)
```
mcrutch: This redirection is what is causing the javascript problem. digging a bit more

mcrutch:
@manfromjupyter Tag, you're it.  Reverted your javascript hack and fixed the underlying problem.
How I found it:
Breakpoint in your javascript error function, checked network tab, checked response headers from the initial PATCH action.  Saw the "Location" header was sent back which causes the redirect to the other page.  Instead, now render a JSON response and let the javascript do the refresh.

mcrutch:
Caught this back up to the main branch.  I want to hammer this out so we can stop adding turbolinks stuff to the codebase (like federated search did).  @manfromjupyter try and hit this with some regression testing, I will knock out testing the federated search stuff

PR #145, opened 2020-03-02
Cron schedules in config and ability to disable sunpot in dev

chris55237:
Cron schedules are defined in config instead of being hard coded. Added ability to enable / disable sunpot in development environment


config/environments/development.rb
```diff
@@ -14,7 +14,7 @@
   config.action_controller.perform_caching = false
 
   config.action_mailer.perform_deliveries = true
-  config.action_mailer.raise_delivery_errors = true
```
chris55237: Removed this as it's also defined on line 36 below

PR #146, opened 2020-03-02
Ui changes

chris55237:
A few changes in this pull request:
- Ability to allow users to propose their own reviews - can be enabled / disabled by config value. By default, it is disabled
- Show gray badge when a previous revision of a notebook has been fully reviewed (and the current revision hasn't)
- Ability to customize stripping of notebook data in extensions
- Metadata endpoint for revisions
- A few bug fixes: users couldn't change group membership, no error was shown if the editing of notebook descriptions failed and CSS padding on modals fix

### DIFF
--- a/app/views/notebooks/_notebook_info_jumbotron.slim
+++ b/app/views/notebooks/_notebook_info_jumbotron.slim
@@ -120,7 +120,7 @@ div.content-container
                   a.modal-activate href="#editNotebookModal" id="editNotebook" data-toggle="modal"
                     span.gear-dropdown-icon.glyphicon.glyphicon-upload aria-hidden="true"
                     | Upload new version
-                -if GalleryConfig.user_permissions.propose_review
+                -if GalleryConfig.user_permissions.propose_review && !@user.admin?
                   li
                     a.modal-activate href="#proposeReviewModal" id="userProposeReview" data-toggle="modal"
                       span.gear-dropdown-icon.fa.fa-file-code-o aria-hidden="true"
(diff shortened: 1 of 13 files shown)


app/views/notebooks/_notebook_info_jumbotron.slim
```diff
@@ -120,13 +120,18 @@ div.content-container
...
+                      span.gear-dropdown-icon.fa.fa-file-code-o aria-hidden="true"
+                      | Propose for Review
                 li
                   a href="#" id="deleteNotebook"
                     span.gear-dropdown-icon.glyphicon.glyphicon-trash aria-hidden="true"
                     | Delete notebook
                     span id="deleteConfirm"
                     span id="deleteWaiting"
-              -if @user.admin?
+              -if !GalleryConfig.user_permissions.propose_review && @user.admin?
```
mcrutch: My concern here is that it is marked as the admin section and other things might get thrown in here (like deprecation) and this won't even create the admin section now if users are allowed to propose their own reviews.

chris55237: Good point - maybe "Propose for review" always stays in the admin section for admins, and only appears in the user menu if propose_review is enabled and the user is not an admin?

mcrutch: Probably for the best.

chris55237: I've updated this now

PR #150, opened 2020-03-05
Added placeholders for potential extensions in a couple places

manfromjupyter:
- Added possible extensions for on default sorting the /reviews and /change_requests/all tables
- Added possible extension to add a label to top of notebook listings
- Made /reviews and /change_requests/all tables consistent with one another
  - Rendered lock icons next to private notebooks on the /change-requests/all page so users know that potentially important piece of information

Tested that the table sorts by default by title and tested a lot of different potential characters to ensure sorting is working


manfromjupyter:
@mcrutch, was this fine or did you want me to look into running a configurable/environmental method?

app/views/change_requests/all.slim
```diff
@@ -12,10 +12,20 @@ div.content-container.center
           th Status
       tbody
         -sorted_array = @change_requests.sort {|a,b| a.notebook.title.partition(')')[2] <=> b.notebook.title.partition(')')[2] }
```
mcrutch: *ahem*

manfromjupyter: Oh wow, I don't know how I didn't see that

mcrutch:
I see table_nb_title referenced, but no matching file. How is that black magic working?

manfromjupyter:
@mcrutch, it won't let me comment on your comment above for some reason. I merely moved that file from one place to another to make it alphabetic order in this ticket. That partial has existent in the past already and merely is the title rendering for notebook listings.

PR #157, opened 2020-03-09
Rather than directly use DeprecatedNotebook controller, use object me…

mcrutch:
…thods on Notebook

I believe everywhere that deprecated can appear is fully tested, except the change request screen which is broken for my user due to corrupt entries in the database.


manfromjupyter:
Looks good to me/ I made some minor changes and added 2 minor fixes to deprecated notebooks to prevent duplicates from being added and empties he notebook autocomplete form upon selecting one of the dropdown items. Tested all of it.. If you approve of my changes I can close it up.

PR #163, opened 2020-03-13
Federated search

chris55237:
First draft of federated search - users will see no change unless federated galleries are added to the config. In this case, all search becomes ajax driven so navigation between the different gallery search results can be done on the same page.

On initialization, each gallery is checked for connectivity with an ajax call to /notebooks.json. If this fails, then the gallery is marked as disabled and the user is unable to select that view that gallery's results. I did consider removing the tab altogether but this creates a scenario where the user sees tabs randomly disappear when the ajax call fails.


app/assets/javascripts/gallery.js.erb
```diff
@@ -179,19 +179,19 @@ document.addEventListener("turbolinks:load", function(){
   /* ======= Home Page Functionality ======= */
   /* ======================================= */
   $(document).ready(function(){
-    var selector = '.tabs a'
+    var selector = '.page-home .tabs a'
     $(selector).on('click', function(){
-      $(selector).removeClass('active');
+      $(selector).closest("li.tab").removeClass('active');
```
chris55237: This is changing the home tabs to make the HTML classes consistent with bootstrap's way of dealing with tabs

docs/configuration.md
```diff
@@ -45,3 +45,24 @@ Admin users can then modify other user accounts from the `/users` endpoint, avai
...
+```
+
+The external gallery will need to accept cross origin requests from the connecting gallery. This can be done in the external gallery's configuration
+as follows:
+```yaml
+search:
+  allowed_cors:
+    - myurl.com
+```
+This value may be a regex or '\*' to allow all origins.
```
mcrutch, line 68: Since we are using credentials, '*' is not a valid origin per the specification (took me a a bit to figure out why this was throwing an error while testing)

Issue #47, opened 2019-06-16
Admin user creation must happen after migrations


jeffyjefflabs:
The code added in a8f9f3e0 needs to move.  In its current location it will run before initial migrations, causing the error in #45.  I'll have to think about the right place for it.  Maybe it should just be a rake task.

mcrutch:
There is now a script to create a user and there was already a script to create an admin user from an existing user.  Not pretty but it works as an alternative.

PR #167, opened 2020-04-29
Document minimum password length

tubaman:
Without setting a correct password with the minimum length, admin user login fails with a 500.


jeffyjefflabs:
Thanks!

For reference, password length requirements are configured [here].(https://github.com/nbgallery/nbgallery/blob/master/config/initializers/devise.rb#L157)

Also, setting those environment variables to create an admin user may not work during first-time startup; see #47 and #112.

Issue #154, opened 2020-03-06
Pass gallery URL in metadata when executing "Run in Jupyter"


mcrutch:
When sending a Notebook to Jupyter, add one more metadata field to the notebook for the gallery url so that the extensions can be modified to use that field for reporting metrics/health/saving to gallery.  NBGallery Extensions for Jupyter will also be modified to support this field as an override for the extension.
Strip this value off when saved to the gallery.
Solves two problems:
Gallery URL changes
Supporting multiple gallery instances from one Notebook instance.

mcrutch:
@jeffyjefflabs metadata, gallery, gallery_url?

jeffyjefflabs:
Sounds good to me!

PR #172, opened 2020-06-16
Add Gallery_url to Notebook metadata when sending to Jupyter

mcrutch:
I am just using the base_url at this time because that ensures it caputres the URL it actually came from for easiest flexibility. I am open to other ideas on this though.


mcrutch:
Closes #154

Issue #162, opened 2020-03-12
Investigate Lograge (or alternative)


mcrutch:
When you are trying to find slow queries/page loads/etc... in production the default format for logs that rails use are very hard to figure out what page was requested to see what the load time was, and logs get very verbose.  Lograge addresses this.  See if implementing this will make things better.

mcrutch:
at log-level info
Without Lograge
Processing by StaticPagesController#home as HTML
  Rendered static_pages/_custom_home_search_fields.slim (2.2ms)
  Rendered application/_alert_container.slim (4.9ms)
  Rendered static_pages/home.slim within layouts/layout (928.2ms)
  Rendered application/_html_head.slim (143.9ms)
  Rendered application/_custom_banner.slim (2.2ms)
  Rendered application/_custom_buttons.slim (2.1ms)
  Rendered application/_custom_links.slim (2.1ms)
  Rendered modals/_login_modal.slim (43.8ms)
  Rendered application/_tos.slim (4.3ms)
  Rendered application/_custom_fields.slim (2.5ms)
  Rendered application/_tos.slim (0.1ms)
  Rendered modals/_upload.slim (35.0ms)
  Rendered modals/_environment_loading.slim (3.3ms)
  Rendered modals/_new_group.slim (10.7ms)
  Rendered application/_custom_footer.slim (2.1ms)
Completed 200 OK in 1518ms (Views: 1386.4ms | ActiveRecord: 29.7ms | Solr: 0.0ms)
Started GET "/home_notebooks" for 10.112.25.20 at 2020-06-17 10:10:15 +0000
Processing by StaticPagesController#home_notebooks as HTML
  Rendered application/_custom_notebook_listing_label.slim (2.3ms)
  Rendered application/_language_icons.slim (5.9ms)
  Rendered application/_table_nb_title.slim (6.5ms)
  Rendered application/_notebook_run.slim (3.6ms)
  Rendered application/_table_nb_description.slim (6.1ms)
  Rendered application/_notebook_metadata_metrics.slim (63.8ms)
  Rendered application/_nb_stats.slim (82.4ms)
  Rendered application/_time_ago.slim (3.2ms)
  Rendered application/_author_rep_trophy_icon.slim (10.7ms)
  Rendered application/_custom_notebook_listing_label.slim (0.1ms)
  Rendered application/_language_icons.slim (0.7ms)
  Rendered application/_table_nb_title.slim (0.7ms)
[...]

PR #178, opened 2020-06-17
Logging tweaks to make production logs a bit more useful by using Log…

mcrutch:
…rage

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -93,6 +93,8 @@ gem 'redcarpet'
 gem 'rinku'
 gem 'rouge'
 
+#Better logging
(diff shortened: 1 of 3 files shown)


mcrutch:
See #162 for a comparison of what pre/post lograge at the info level look like.  It is a huge difference and worth the trade-off when managing a moderate->high traffic production system.

Is there anything I am overlooking @jeffyjefflabs?  If you turn to debugging. debug messages still get spit through as-is so for those rare cases we turn the logging dial up to 11 we aren't losing those, just the breakdown of individual views which at that point you should be investigating off-prod

PR #195, opened 2020-06-19
Hide the subscribe icon for non-authenticated users

### LINKED ISSUE #184: Remove subscription icon on users, groups, and tags for Anonymous users
Super easy, just need to wrap it in a

```-if @user.member?```

Will do it myself in a short amount of time if nobody does it sooner.

mcrutch:
Closes #184


manfromjupyter:
Tested as anonymous user, anonymous accessibility user, and a signed in user... All work as expected. These changes are good to go

PR #221, opened 2020-07-01
Hiding the table and display an informative message when no environme…

### LINKED ISSUE #205: Don't render environment table unless an environment exists for that user
alternative is can render the column but just have the none-table-head be a single column saying there are no results.

mcrutch:
…nts are found
Closes #205

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -224,6 +224,9 @@ table.clean-table {
         text-align: left;
     }
 }
+td.no-entries {
+    padding: .75em 0;
+}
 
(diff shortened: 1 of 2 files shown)


manfromjupyter:
I know this is exactly what I asked for but it doesn't seem right to me. I have a different idea on how to make it look and will change this pull request with that change when I can spare 30 minutes to do it to see if you like it as well.

manfromjupyter:
Approved if you like my changes.

Will let users know what kind of metadata should be apart of their environments or baout hwat it should look like if it did have entries. Works better than before for accessibility users.

no environments in table

PR #231, opened 2020-07-07
Notebook seeding from the command line

mcrutch:
A partial solution for bulk loading of notebooks


manfromjupyter:
The code looks good by what I saw. In terms of getting it to work, I entered my username twice 2 times and tried entering the path of a folder of notebooks I currently have on my desktop. Is there a particular format the url should be in?

didnt throw error for some reason

PR #238, opened 2020-07-16
Update to actionview/related packages

mcrutch:


mcrutch:
You guys make sure this builds/runs ok for you. Dependabot couldn't do the updates for some reason, but I think it is because 4.2.11.2 was a broken build.

PR #242, opened 2020-07-17
Fix custom_links display for both ajax and mobile menus

mcrutch:


mcrutch:
Finishes off #216 because essentials are hidden from non-logged in users BUT custom links are added to the bottom of the menu in their own section if created.

app/views/layouts/beta_layout.slim
```diff
@@ -56,9 +56,23 @@ html lang="en"
                           div.nav-bar-icon-container
                             div.nav-bar-icon
                       ul.dropdown-menu id="mobileNavDropdown" style="display: none"
-                        -if @user.member?
-                          li.dropdown-header.filter-item Essentials
-                          ==render partial: "custom_links"
+                        -custom_links=render('custom_links')
+                        -if @user.member? || custom_links.length > 0
+                          li.dropdown-header.filter-item Quick Links
+                          -if custom_links.length > 0
```
mcrutch, line 62: Line 62 not needed since you aren't doing a divider anymore and you already checked length.

app/views/layouts/layout.slim
```diff
@@ -56,9 +56,23 @@ html lang="en"
                           div.nav-bar-icon-container
                             div.nav-bar-icon
                       ul.dropdown-menu id="mobileNavDropdown" style="display: none"
-                        -if @user.member?
-                          li.dropdown-header.filter-item Essentials
-                          ==render partial: "custom_links"
+                        -custom_links=render('custom_links')
+                        -if @user.member? || custom_links.length > 0
+                          li.dropdown-header.filter-item Quick Links
+                          -if custom_links.length > 0
```
mcrutch, line 62: Not needed

app/views/static_pages/layout_dropdown.slim
```diff
@@ -1,5 +1,8 @@
--if @user.member?
-  li.dropdown-header.filter-item Personal
+-custom_links=render('custom_links')
+-if @user.member? || custom_links.length > 0
+  li.dropdown-header.filter-item Quick Links
+  -if custom_links.length > 0
```
mcrutch, line 4: Not Needed

app/views/layouts/beta_layout.slim
manfromjupyter: It could be a signed in user on site without custom links. We still would want the Quick Links section to render, just not the subsection within it that reveals custom_links.

app/views/layouts/layout.slim
manfromjupyter: Same as above unless I'm not seeing something here

PR #248, opened 2020-07-27
Permissions Fix: Removed owner actions from users with editor permissions

### LINKED ISSUE #258: Only allow notebook or group owners to change ownership, delete, toggle private/public and share notebooks

manfromjupyter:
Closes #258

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -546,6 +546,15 @@ class ApplicationController < ActionController::Base
     verify_edit(false)
   end
 
+  def verify_owner
+    type = Notebook.find(@notebook.id).owner_type
+    if (type == "User" && @notebook.owner_id != @user.id)
+      raise User::Forbidden, 'Restricted to users with owner permissions.' unless @user.admin?
+    elsif (type == "Group" && Groups_User.where(user_id: @user.id, group_id: Group.find(@notebook.owner_id)).owner)
+      raise User::Forbidden, 'Restricted to users with owner permissions within the group.' unless @user.admin?
+    end
+  end
+
   # Get the staged notebook
   def set_stage
     @stage = Stage.find_by!(uuid: params[:staging_id])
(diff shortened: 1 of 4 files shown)


mcrutch:
Still poking at this but might need @jeffyjefflabs to chime in

mcrutch:
Nevermind, found your problem, you aren't blocked anymore.

manfromjupyter:
Oh hah, fail. Okay will finish it right now. Thank you!

app/models/user.rb
```diff
@@ -262,6 +262,17 @@ def can_read?(notebook, use_admin=false)
       (use_admin && admin?)
   end
 
+  def owner(notebook)
```
mcrutch, line 265: You wrote this function...but then don't use it in the verify_owner function?

manfromjupyter: I used it for the view. I guess I forgot I might be able to use it for the controller. Will play with it in a few

manfromjupyter: Got it now. Works perfectly. Good eye

app/controllers/application_controller.rb
```diff
@@ -546,6 +546,10 @@ def verify_edit_not_admin
     verify_edit(false)
   end
 
+  def verify_owner
+    raise User::Forbidden, 'Restricted to users with owner permissions.' unless @user.owner(@notebook)
```
mcrutch, line 550: Should also allow admin here

manfromjupyter: The method in the user model covers admin there already. 
Tested it and it checks out.

```def owner(notebook)
    if notebook != nil
      type = Notebook.find(notebook.id).owner_type
      if ((type == "User" && notebook.owner_id == id) || (type == "Group" && GroupMembership.where(user_id: id, group_id: Group.find(notebook.owner_id)).pluck(:owner)[0]) || admin?)
        return true
      end
    else
      return false
    end
  end```

Issue #249, opened 2020-07-28
Auth not working


cccs-cat001:
Hi, I have NBGallery set up. It has an email address and db and all that and it all seems to be working. When I go to register, nothing happens it remains on the register screen. I check the db and my user info is there, so something happened. And I never receive an email from nbgallery asking me to confirm my email. 

Can someone help me debug this, I'm not a ruby dev by any means...

Also is there a way to remove auth and just have it open to everyone?

mcrutch:
There isn't any way to remove auth entirely because user objects are needed to relate to the other objects.   

The problem most likely stems from the fact that the email server is not properly configured so the confirmation email never gets sent so the account isn't activated.  For testing purposes until you can get email configuration sorted out you can set users.confirmed_at to the current date and users.approved=1 for the user in the mysql database. That should do it. 

I will say the EASIEST way to get auth working long-term is to use one (or more) of the oauth provider options, gitlab, github, google, or facebook, that are available.  

See: https://github.com/nbgallery/nbgallery/blob/master/docs/configuration.md for information about the email configuration and the oauth documentation. (Although I have to add the link to the gitlab side of the documentation, forgot that)

mcrutch:
Oh, btw, email configuration is required for some of the features to work so you have to get that sorted out anyway (Change requests, feedback, shares) but for development purposes I have been using docker image gessnerfl/fake-smtp-server and forcing it to listen on port 587.

  smtp:
    image: gessnerfl/fake-smtp-server:latest
    ports:
      - "5080:5080"
      - "5081:5081"
      - "587:5025"
    restart: always
THen passing the appropriate EMAIL_SERVER, EMAIL_DOMAIN, GALLERY__EMAIL__GENERAL_FROM and EMAIL_DEFAULT_URL_OIPTIONS_HOST environment variables to the docker container.  I will write this up in more detail sometime.

cccs-cat001:
See, I'd like to use oauth, but we use azure AD, which isn't listed for nbgallery. The email server I have set up is just a relay listening on port 25. Would there be any issue with changing the port to be 25 instead of 587?

mcrutch:
There is a ticket for doing a sample ldap authentication plugin, but it hasn't bubbled to the top of the stack.  

I also opened #250 to allow configuring the email port.

cccs-cat001:
Cool thanks

cccs-cat001:
[...]

PR #252, opened 2020-07-28
Email port configuration

### LINKED ISSUE #250: Allow configuration of EMAIL_PORT
This is something that has bitten me before, need to allow other ports for non-standard configurations

mcrutch:
closes #250.  Had to account for both empty string AND unset value and the ENV['EMAIL_PORT'] || 587 only accounted for unset, not empty string.


mcrutch:
Made some additional changes to try and solve the #249 ticket so hold off on merging just yet.

mcrutch:
closes #249

PR #257, opened 2020-07-30
Update SOLR container to 8.x in docker compose

### LINKED ISSUE #55: docker can't reliably mount files as volumes unless the file already exists in the container
While using docker-compose, solr would not stand up because /opt/solr/server/solr/solr.xml does not exist in the official solr container.  Consequently, it throw an error because it auto creates a directory called "/opt/solr/server/solr/solr.xml" and then tries to mount a file there as a volume.

ref: https://docs.docker.com/storage/bind-mounts/
"If you use -v or --volume to bind-mount a file or directory that does not yet exist on the Docker host, -v creates the endpoint for you. It is always created as a directory"

I have seen inconsistent behavior between different versions of docker, but docker 18 and docker-compose 1.24 didn't work.

mcrutch:
and provide a more reliable way to mount the configuration to resolve #55


mcrutch:
While working with a user (#263), tested this in a completely clean environment (never having had NBGallery before) and it worked, but I have tweaked the documentation to put the solr permissions issues front and center giving the commands that will resolve it before you start solr the first time, so merging this in.

PR #270, opened 2020-08-11
Allow disabling of username login and registration

### LINKED ISSUE #267: Ability to turn off account registration and internal login capability
In otherwords, make it possible to support oauth only, but break registration and login into two different fields so you could still use internal login but not allow people to "register" so only admins can create accounts.

mcrutch:
Closes #267 
This doesn't disable the underlying devise functionality because that would screw with some schema stuff that we don't want changing based on configuration, but disables all the views that enable username login, password resets, user registration, and resending confirmation based on configuration.


app/views/devise/shared/_oauthlogin.html.slim
```diff
@@ -0,0 +1,18 @@
+- if devise_mapping.omniauthable?
```
mcrutch, line 1: @manfromjupyter Title and screen reader only text is probably overkill here, but I put in both.  I want something for the sighted user in a hover-text, so what do you think is best here.

app/assets/stylesheets/modals.scss
```diff
@@ -391,25 +391,27 @@ table .form-group {
...
-        background-color: #3b5998;
-    }
-    .github {
-        background-color: #444;
-    }
-    .gitlab {
-        background-color: #3f3177;
-    }
+}
+.oauth-login-container{
```
mcrutch, line 395: Just allowing these styles to be used outside of just the modal if you go to <url>/users/sign_in

app/views/devise/shared/_oauthlogin.html.slim
manfromjupyter: Title doesn't help most if not any screenreaders in my experience. Some are smarter than others, but for JAWS, the most popular one of them all, the way you have it is perfect.

app/views/devise/registrations/new.html.slim
```diff
@@ -1,27 +1,28 @@
...
-          br
-          = f.password_field :password, autocomplete: "off"
-        .field
-          = f.label :password_confirmation
-          br
-          = f.password_field :password_confirmation, autocomplete: "off"
-        .actions
-          = f.submit "Sign up"
-      = render "devise/shared/links"
+-if GalleryConfig.registration_allowed == true && GalleryConfig.username_login_allowed == true
```
manfromjupyter, line 1: @mcrutch, When I turn the `registration_allowed` to `false` in the configs/settings, when going to the url: `SITE/users/sign_up`, the page isn't throwing a 404 or a page with an error on it, just showing up as a blank page. Nothing is exposed in the inspect element for this and directing to this page and the confirmation new, but just war curious if we wanted to do that or keep it as is?

jeffyjefflabs: I tried something similar.  I set `username_login_allowed: false` and `registration_allowed: true` - does that even make sense as a valid configuration?  The views seemed weird at first (basically nothing except the github oauth button), but maybe it's actually doing the right thing.  Maybe it at least needs some words about how to get an account in that scenario?

jeffyjefflabs: But yes other than that it looks ok to me.

mcrutch: We could add some placeholder text to those views, but since those routes are built into devise it can get a little complicated to disable them in a more "error page" kind of way based on some of the things I saw on the web.

username_login_allowed:false automatically blocks out registration too.  Anyone with an account on your oauth provider can get in that way even if they don't already have an account but that was kind of my plan with this and it would be made clear in the docs.  Prevent people from signing up/logging in without using oauth, or just prevent random people from signing up for accounts and have it all done by admins on the backend. (The latter is the more likely scenario we are going to need soon)

jeffyjefflabs: > username_login_allowed:false automatically blocks out registration too

Ok I thought that might be the case - makes sense.  I wouldn't worry too much about the views then.

manfromjupyter: @mcrutch apologies, got distracted with the orgs ticket, will make the changes asap tomorrow for the alerts on those blank pages, then will approve it.

mcrutch: @manfromjupyter I am making a few tweaks...since most of them render devise/shared/links at the bottom, as long as you have an oauth provider enabled something will show up (which would be the intended behavior for username_login disabled.  for sign_up I am tweaking just a bit.

PR #274, opened 2020-08-13
Update packages form the core ruby 2.3 image to deal with some CVEs

mcrutch:
Been running a variant of this in development.  Do not merge yet, I will spend some time with the docker images built from this branch.


mcrutch:
This causes the image to grow 130MB but resolves a bunch of CVEs (15 critical, 566 total)

Issue #194, opened 2020-06-18
Homepage listings reload when notebooks dropdown in header expands


manfromjupyter:
Fix how homepage listings reload when notebooks dropdown in header expands

mcrutch:
Do you mean the drop-down reloading when you expand it?  Because, yes, it shouldn't reload every time but we removed it from the default page load because it is super slow.

PR #289, opened 2020-09-02
Prevent the dropdown menu from being loaded repeatedly on the same pa…

mcrutch:
…ge load

Related to #194.  Not sure if it was the problem @manfromjupyter was talking about, but it was a problem and I couldn't replicate what was described in the ticket.

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -514,14 +514,19 @@ $(document).ready(function(){
   if(pathname.indexOf("stage") > -1){
     var staging_id = pathname.substr(pathname.lastIndexOf('/')+1);
   }
-
+  var dropdown_loaded = false;
   $('#dropdownCaretButton').on('click',function(){
-    $('#dropdownMenu').html('<center> <%=image_tag('loading.gif', style: 'height: 20px') %> </center>');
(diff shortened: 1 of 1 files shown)


manfromjupyter (commented):
These changes do seem to address the issue that needs to be fixed for that ticket, but one thing is minorly off. If you load the home page, select a different notebook listing tab, wait for it to load, and then select the notebook dropdown above, the nbgallery loading logo still appears and pulse for a second or two, but after toggling there a second time, there is no longer a delay in the notebook dropdown revealing because it just skips the loading gif now within it because that content has already loaded (thank you) AND the nbgallery loading logo doesn't appear anymore.

Maybe it's beyond the scope of the ticket, but I think we need to rename the loading gifs so that they are titled differently or increase specificity in the javascript that calls them to fix both the issue this ticket fixes and that other one mentioned in the original ticket.

mcrutch:
I fixed the loading icon issue as well in my last commit.  The danger in the way it was is that that logo would fire ANY time an ajax request is going on, so I made it unregister itself after it finishes with the current ajax request.

manfromjupyter (approved):
Clever solution with the JS unbind. Didn't even know that was a thing hahahah. Approved

PR #292, opened 2020-09-03
Added the ability to filter the "All Users" page used by admins

mcrutch:
Closes 122
Could use an accessibility check-over by @manfromjupyter because I haven't found a good document for "smart tables" for filtering records.

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -2510,6 +2510,11 @@ body.page-summary .form-group {
     }
 }
 
+/* ===== Users Page (/users) ===== */
+body.page-users table th .form-control {
+    padding-right: 0 !important;
+}
+
 /* ===== Edit and New user Form ===== */
 #userForm {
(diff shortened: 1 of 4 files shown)


manfromjupyter (approved):
Approved. I made a couple changes to it though.
1. Added the form attribute to the button so the browser knows what form that submit button belong to.
2. Removed the searchFieldBox class to the first 4 table column header inputs so that selecting them doesn't cause the Header search input's info icon to appear and then disappear.
3. Added aria-labels to the inputs for accessibility users because without it they don't know what these inputs do (there is no mention of filtering anywhere else)
4. Worked around the bootstrap auto-injection of padding-right on each of the input fields that gets added when the field is focused which causes the table to shake in that is causes the column to grow a couple pixels and causes all others to get a little smaller. Now selecting those filter fields doesn't cause table to move.
5. Changes mentions of "user" to "username" so it more accurately describes it/reflects how we name it in our database/etc

PR #294, opened 2020-09-14
Resource links

### LINKED ISSUE #153: Metadata field for additional resources (0..n links) for Notebooks
Allow attaching 0..n additional links of external references to notebooks
Default Title of the field/section is "External Resources" but will be stored in a configuration variable
provide hooks for pre-save and pre-display in case additional sanitization is required.
Index in solr
Provide search filter

mcrutch:
@manfromjupyter Take a look through this.  The "External Links" section only appears if
a) There are links or
b) The user viewing the notebook has edit priv (including admin) so they have access to the + icon

Dialog copies error message method from a different dialog, so should be accessible, unless I copied one that wasn't.

Simplistic Javascript validation and more robust validation for URL on the backend
Doesn't prevent double posting right now but, they are easy to delete...What if they want two titles to the same doc or something.

Closes #153

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -10,8 +10,8 @@ $(document).ready(function(){
       $('#mobileNavContainer').click();
     }
   });
-  $('#mobileNavContainer').click(function(event) {
-    event.preventDefault();
+  $('#mobileNavContainer').click(function(e) {
+    e.preventDefault();
     $(this).toggleClass('expand');
     $('#mobileNavDropdown').toggleClass('open');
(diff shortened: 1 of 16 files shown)


manfromjupyter:
@mcrutch I adding some accessibility functionality to this and editing some of it for consistency with the rest of the product. I am almost one, just need to change the styling of the delete icon so that tabbing to it reveals it (if you support it being hidden by default until hovered over the element; can talk more later

manfromjupyter (approved):
Might want to take one more look at it to see if you like it, but otherwise all approved by me. :)

PR #295, opened 2020-09-15
Visibility/execution of disabled notebooks

### LINKED ISSUE #196: Hide disabled notebooks from Jupyter starred and recently executed notebooks lists
Back in March, 2020 a feature was added to deprecated and disable notebooks. Upon disabling the notebook is not capable of being run but users can still navigate to the direct URL in Jupyter to run it. If there is a way we can prevent this from NBGallery's source code, do it here so if the environment changes or varies with other user's project, it will always work. May not be plausible but worth checking out

EDIT: Find way to prevent disabled notebooks from showing up in Jupyter's Starred notebooks or Recently Executed notebooks

### LINKED ISSUE #311: Deprecation status changes do not get updated in Solr
if you deprecate/undeprecate a notebook, the SOLR record is not getting reindexed so until a total reindex happens the status is wrong.

mcrutch:
A few parts to this one.
1 - Deprecated are no longer sent to Jupyter automatically from the "Recently Executed" and "Starred" (Since we don't allow "Run in Jupyter, why would we auto load them)  Because of how these are done it meant hiding them from the view by default so...
2 - Deprecated no longer appear in any view by default (Group list of notebooks, search, my stars, shared with me, etc...) but there is a checkbox to toggle them on.
3 - On the homepage there won't be a toggle to show them, which I am ok with. they were deprecated for a reason

I copied the current functionality and labeling of the sort dialog so accessibility should be ok, but need @manfromjupyter to just sanity check that.

Closes #196 
Closes #311


jeffyjefflabs (approved):
I didn't run it live but looks ok to me!

manfromjupyter:
@mcrutch Looks to be 1 maybe 2 issues.

It works in that deprecated notebooks that have a frozen state on them (too unsafe to run) aren't being populated in those Jupyter folders but the deprecated notebooks that are still safe to run just abandoned or obsolete also aren't appearing in there. Not sure if this is an issue but I think it is. If so, just need to instead do a check for ```if deprecated_notebook.disable_usage``` (might need a "== 1"). 

Issue I'm encountering is that I can't search without getting the SOLR issue shown. I tried restarting my dev server too.
solr issue

mcrutch:
What were you doing when you got the above error?

mcrutch:
Ok, search is fixed.
As for the deprecated/frozen - This is the problem with an overloaded feature.  For the UI purposes not showing any deprecated notebook by default makes the most sense, and the extensions in Jupyter just use the same exact endpoints as the UI.  Two different filters would get weird for that since it is "Deprecated" and "Deprecated with Special Sauce"

I think the appropriate action is to not load any deprecated notebooks by default because that draws attention to the fact that the notebook is deprecated and they should try to look elseware.  Sure, they can manually load those notebooks still, but it is not an automagic thing anymore and might be the only way for them to find out that the notebook is now deprecated.

manfromjupyter:
@mcrutch doing a search for nothing (tested with an anonymous user) returns Sunspot error still (url: http://{IP_ADDRESS}/notebooks?q=&sort=score

mcrutch:
Grr...fixed.  Also, found a bug to address later, but Notebooks are not getting re-indexed in solr when they deprecate/un-deprecate

manfromjupyter (approved):
Search works as expected now. Approved.

Wanted to note one thing though: I have not been much on this page in a while but I never realized it uses a lot of camelcased classes which we don't do typically; it's mostly been that ids get camelcased and classes get dashed for the HTML/CSS. I made a minor change or two to this to clear up like 15 of them but ensure it still works, but I want to and will eventually go back to make sure the rest is working, but can't test them all due to Tags and Groups search results not appearing from said searches (ex. search "panda"). Anyway, all looks good, just wanted to explain what I did in addition to a minor css centering to the positioning of the deprecation checkmark.

mcrutch:
" but I want to and will eventually go back to make sure the rest is working"
Either sort out your testing or roll back the changes.

mcrutch:
Nevermind, tested and merged.

manfromjupyter:
Apologies, yes. Meant I will go back later and make the rest follows the same labeling guide format. *It all works :)

PR #313, opened 2020-09-29
Change devise alert and notice to error and info for proper display

### LINKED ISSUE #309: Invalid Password or Email should Return Error Message not Info Message
.alert.alert-error not .alert.alert-alert

invalid password or email should return error message not info

mcrutch:
Closes #309 

This is built into devise (it only returns notice and alert) and the only way to fix this is a little hacky - https://stackoverflow.com/questions/32444342/devise-flash-message-being-notice


manfromjupyter (approved):
Works!

PR #325, opened 2020-10-14
Change Request approvals, declining, and cancelling no longer hangs or returns server errors

### LINKED ISSUE #320: Change Request approvals, cancelling, and declining cause server errors

manfromjupyter:
Closes #320


manfromjupyter:
Just reconfirmed, this is exactly what has changed. I think it might seem like more because this was 2 different hotfixes rolled into one and the first one, just the JS one took me two times to get it right (that one line) so might seem like more.

If we get that email idea thing you were thinking about working, we can test it to make sure. Works when I comment out a couple lines here and there and I verified it is exactly the same change (letter for letter), just can't confirm 100% because of the email thing.

Issue #323, opened 2020-10-10
Deprecation Reasoning column is t.string but should be t.text


manfromjupyter:
When including a reasoning for deprecation that is longer than a given amount of characters, it 500 errors for being too long. Need to 1.) make the database variable change and then 2.) error catch for size of the reasoning. Limit to idk 10000 characters (through ruby)

PR #329, opened 2020-10-14
GITHUB-323: Deprecated Notebook's SQL table reasoning column no longer return Errors

manfromjupyter:
Switched Deprecated Notebook's SQL table reasoning column from t.string to t.text so will no longer return rails/MYSQL error: 'ActiveRecord::StatementInvalid - Data too long for column'

Test by either inspecting deprecate_notebooks table or adding a huge amount of text to a deprecation status.

Closes: #323

### DIFF
--- a/db/migrate/20201014151211_change_reasoning_to_be_text_in_deprecated_notebooks.rb
+++ b/db/migrate/20201014151211_change_reasoning_to_be_text_in_deprecated_notebooks.rb
@@ -1,0 +1,9 @@
+class ChangeReasoningToBeTextInDeprecatedNotebooks < ActiveRecord::Migration
+  def up
+    change_column :deprecated_notebooks, :reasoning, :text
+  end
+
+  def down
+    change_column :deprecated_notebooks, :reasoning, :string
+  end
+end
--- a/app/assets/javascripts/notebook_actions.js.erb
+++ b/app/assets/javascripts/notebook_actions.js.erb
@@ -568,6 +568,23 @@ $(document).ready(function(){
     });
     return false;
   });
+  $('#deprecateNotebookReasoning').keyup(function(){
+    maxlength = 500;
+    length = this.value.length;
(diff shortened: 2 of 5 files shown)


mcrutch:
Should this also come with some kind of reasonable limit for the length?  That field in the database can now store 2GB of plain text and there isn't any limiting mechanism for it.
image

db/migrate/20201014151211_change_reasoning_to_be_text_in_deprecated_notebooks.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeReasoningToBeTextInDeprecatedNotebooks < ActiveRecord::Migration
+  def up
+    change_column :deprecated_notebooks, :reasoning, :text
```
mcrutch, line 3: Went back and looked at the ticket.  Ticket talks about restricting length in Ruby.  This can't be merged without the corresponding piece otherwise the codebase is in a state where extremely large values can make it into the database (and once this is merged, it becomes part of the docker image).

mcrutch (changes requested):

db/migrate/20201014151211_change_reasoning_to_be_text_in_deprecated_notebooks.rb
manfromjupyter: Oh shoot, I totally forgot about that. I agree, you're right, we definitely need a limit and error handling that enforces it. I'm on it.

manfromjupyter:
@mcrutch, it should be good now. Added front end max character count of 500 characters that can entered in the input field and included a character count that appears below the form when you have 50 or less characters to use. At 500/500 characters it goes bold and red. And to prevent people trying to break it, did error handling from server side as well. Returns just a flash message with said error that it was too long. Was going to do the error inline and not refresh the page, but since this can only happen if they're trying to break it, hack it, etc, I thought this was the ideal solution.

mcrutch:
Merge conflict, both changes are yours, so I will let you resolve the conflict.  Otherwise happy to merge once that is complete.

PR #338, opened 2020-10-22
GITHUB-334: Update docker-compose YAML to use more recent base-image instead of Alpine

### LINKED ISSUE #334: Update docker-compose-with-jupyter.yml to use base-notebook image
docker-compose-with-jupyter.yml currently uses an alpine Jupyter image and needs to be converted to use the nbgallery/base-notebook image.

nklissus:
This should be a very simple swap and everything seems to work with the base-image just swapped in. This should fix #334

### DIFF
--- a/docker-compose-with-jupyter.yml
+++ b/docker-compose-with-jupyter.yml
@@ -2,13 +2,12 @@ version: "3"
 
 services:
   jupyter:
-    image: nbgallery/jupyter-alpine
+    image: nbgallery/base-notebook
     ports:
-      - "80:80"
+      - "80:8888"
     environment:
       NBGALLERY_URL: http://localhost:3000
       NBGALLERY_CONFIG_TOKEN: ${NBGALLERY_CONFIG_TOKEN}
       NBGALLERY_CONFIG_PASSWORD: ${NBGALLERY_CONFIG_PASSWORD}
       NBGALLERY_ENABLE_INSTRUMENTATION: 1
       NBGALLERY_ENABLE_AUTODOWNLOAD: 1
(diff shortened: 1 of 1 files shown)


mcrutch:
@jeffyjefflabs Should we just make the Jupyter container run on 8888 since that is the default anyway and if they look at the logs to get the token, they will just be able to grab that URL rather than have to change the port to 80?

docker-compose-with-jupyter.yml
```diff
@@ -2,13 +2,12 @@ version: "3"
...
 services:
   jupyter:
-    image: nbgallery/jupyter-alpine
+    image: nbgallery/base-notebook
     ports:
-      - "80:80"
+      - "80:8888"
     environment:
       NBGALLERY_URL: http://localhost:3000
       NBGALLERY_CONFIG_TOKEN: ${NBGALLERY_CONFIG_TOKEN}
```
mcrutch, line 10: base-notebook does not support NBGALLERY_CONFIG_TOKEN or NBGALLERY_CONFIG_PASSWORD.  @jeffyjefflabs said he might ad support for it in the future (https://github.com/nbgallery/docker-images/issues/8#issuecomment-531998137) but it is not a current issue, so should probably just remove these from the docker-compose if we are using base-notebook.

jeffyjefflabs: Agreed

jeffyjefflabs:
> @jeffyjefflabs Should we just make the Jupyter container run on 8888 since that is the default anyway and if they look at the logs to get the token, they will just be able to grab that URL rather than have to change the port to 80?

Yeah, good catch, I like that idea.

PR #346, opened 2020-11-02
This prevents a redirect loop that attempts to keep submitting the ch…

mcrutch:
…ange request after it is submitted when it originates from Jupyter


mcrutch:
Pull #325 didn't work when uploading a change request from Jupyter because the ?staged=blah parameters were still there.
this method is = window.location.reload without the query string.

manfromjupyter (approved):
Can't test because lacking Jupyter atm but looks right and have confirmed it doesn't brake current change requests just within NBG. Approved if it works for you

PR #179, opened 2020-06-18
Closes #44

mcrutch:

PR #347, opened 2020-11-06
Rollback change that broke Firefox "Run in Jupyter" in client certifi…

mcrutch:
…cate environments


mcrutch:
This rolls back the change from #179

PR #348, opened 2020-11-06
Closes 317 - Sorting Shared with Me

### LINKED ISSUE #317: Shared with Me Notebooks are not sortable
Going to /notebooks/shared_with_me and trying to sort the notebooks does not actually make them re-adjust. May or may not be possible based on how we query the notebooks, but if not, should just hide the sorting button for this page.

mcrutch:
Used a variation of the code from the readable megajoin
to allow sorting of the shared with me page.

Not the most elegant solution, but the readable mega join isn't designed to use the shares relationship.

### DIFF
--- a/app/controllers/notebooks_controller.rb
+++ b/app/controllers/notebooks_controller.rb
@@ -672,8 +672,26 @@ class NotebooksController < ApplicationController
 
   # GET /notebooks/shared_with_me
   def shared_with_me
-    @notebooks = @user.shares.paginate(page: @page)
+    score_str = '(IF(SUM(score), SUM(score), 0.0) + trendiness + IF(review, review, 0.0) + ' \
+      'IF(health IS NOT NULL, 2.0 * health - 1.0, 0.0)) AS score'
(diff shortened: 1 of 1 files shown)


manfromjupyter (commented):
On the shared with me page, I am unfortunately unable, even as an admin, to see private notebooks that are shared with me. As is on our production system, admins and non-admins can see private pages shared with them and pretty sure they need to be able to do this so would like to request that gets included

mcrutch:
That is strange since I am just adding additional joins/wheres to the the existing shares.  Will look into it.

mcrutch:
are you sure it isn't just a deprecated notebook that is missing and you needed to toggle show deprecated?  I even shared a private notebook with myself from a test user and here is what the result was:
image

manfromjupyter (approved):
Woops yeah, you were right, my notebook was deprecated so I thought no private notebooks worked with it. This works splendidly. "Apppppprooooved!" *said in a female cockney accent

PR #350, opened 2020-11-09
Jupyter Classic vs Jupyter Lab UI for "Run"

### LINKED ISSUE #279: If the user visits the /lab url to register, send them to lab when they open the notebook.
URL is of the format http://{server}/{path-to-juptyer}/lab/tree/name-of-notebook.ipynb
Related code that opens the window: https://github.com/nbgallery/nbgallery/blob/master/app/assets/javascripts/run_in_jupyter.js.erb#L155

(Depends on appropriate extensions being in lab to interface with NBGallery)

mcrutch:
Allow extensions to set which Jupyter UI notebooks should open in
Allow users to specify what UI they want their "Run in Jupyter" button to use
Compatible with jupyterlab-nbgallery 0.2.0a3 and above Jupyterlab extensions
Closes #279 

Note: Allows for us to change the default in the future for users who haven't changed the value for their "classic" environments (because those still default to NULL from the extension, but can be manually forced to classic by editing the environment)
Will not change it just because they visit Lab (only new environments created as a result of going to Lab are set to lab)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .env
 .bundle
 /vendor/bundle
+/vendor/cache
 *.key
 *.pem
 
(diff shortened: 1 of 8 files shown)


manfromjupyter (approved):
Approved. Works perfectly. Super cool too

PR #351, opened 2020-11-10
GITHUB-203: Added more context to screenreader text for the home link so screenreader users know what they're clicking on/looking at

### LINKED ISSUE #203: Home link for screenreaders should say "'NB Gallery home' link" instead of "link 'nb gallery'"

manfromjupyter:
Closes #203


manfromjupyter:
Made capital to have JAWS read as a proper noun, so "nb" reads as "en-bee" not "nib".

If you think we should probably turn this into a new environment variable and use like this instead, just let me know.

PR #358, opened 2020-11-18
Allow users who are part of a group to be deleted

### LINKED ISSUE #357: Cannot delete user with group membership: mysql2::error: Unknown column 'groups_users.' in 'where clause'
Hi,

I am getting error deleting a user who has a group membership with the following error:
ActiveRecord::StatementInvalid in UsersController#destory 
line: 196 `@viewed_user.destory`
Mysql2::Error: Unknown column 'groups_users.' in 'where clause': DELETE FROM `group_users` WHERE `groups_users`.`` = NULL

I had no issue deleting user without group membership. Had not tried deleting user with notebook.

Thanks.

mcrutch:
Closes #357 
Intermediate tables MUST have singular primary key in order to be used in a cascading delete it seems.

### DIFF
--- a/db/migrate/20201118162416_add_primary_to_groups_users.rb
+++ b/db/migrate/20201118162416_add_primary_to_groups_users.rb
@@ -1,0 +1,5 @@
+class AddPrimaryToGroupsUsers < ActiveRecord::Migration
+  def change
+    add_column :groups_users, :id, :primary_key
+  end
+end


manfromjupyter (commented):
Approved. Removes user from group when user is deleted as expected without causing any problem to the group. I know you're investigating how it handles notebooks, so will reapprove them, but can confirm this part of it.

mcrutch:
The delete user notebook thing is a whole other ticket that I submitted.  (which is similar to the subscription thing to, which I was investigating something related to that when I stumbled on this...lol)
Deletes are hard when testing for deletion isn't done because we don't normally delete.

jeffyjefflabs (commented):
Looks fine to me

Issue #360, opened 2020-11-18
Subscriptions to tags either has to be redone or removed


mcrutch:
If you subscribe to a tag, in the database, in subscribes to a SPECIFIC instance of that tag.

user_id 1 clicked subscribe to tag "test"
It is recorded as a subscription to that tag AS ATTACHED TO notebook 35.
If that tag is removed from notebook 35, or that notebook is deleted, the subscription breaks and the fact that it was tag "test" is lost (it is just pointing to some ID that no longer exsts)
The tag also exists on Notebook 464

MySQL [gallery]> select * from subscriptions where sub_type='tag' and user_id=1;
+----+---------+----------+--------+---------------------+---------------------+
| id | user_id | sub_type | sub_id | created_at          | updated_at          |
+----+---------+----------+--------+---------------------+---------------------+
| 44 |       1 | tag      |    124 | 2020-11-18 17:43:26 | 2020-11-18 17:43:26 |
+----+---------+----------+--------+---------------------+---------------------+

MySQL [gallery]> select * from tags where tag='test';
+-----+---------+------+-------------+---------------------+---------------------+
| id  | user_id | tag  | notebook_id | created_at          | updated_at          |
+-----+---------+------+-------------+---------------------+---------------------+
| 124 |       1 | test |          35 | 2020-03-10 13:42:59 | 2020-03-10 13:42:59 |
| 134 |       1 | test |         464 | 2020-09-03 12:54:23 | 2020-09-03 12:54:23 |
+-----+---------+------+-------------+---------------------+---------------------+

mcrutch:
Ways to fix it.
- add a column to the subscription table and if it is a "tag" type subscription, it is the text of the tag (Easiest, but leads to repetitive text..just like the current tags table..  Not good normalization in the database, but avoids headaches of the second option)
- make another look up table for tags (which, does it only get updated with a one-up id and text if someone clicks subscribe? Then have to lookup if it exists, and if not add it, and then use that id for all subscriptions to that tag)
- profit?   (seriously, can't think of a third option here right now but there might be better solutions)

jeffyjefflabs:
I like option 2 to fix/normalize the tags table - more work yes, but from a maintenance perspective I think it's good to force things back into "Rails conventional" where we deviated originally.  I think we made this table unconventional because we thought we wanted to track which user had added each tag, but we never did anything with that plus you lose it if the tag is removed anyway

[...]

PR #361, opened 2020-11-18
Address some issues with Notebooks having trouble displaying after a …

### LINKED ISSUE #359: If you delete a notebook updated by a user, you lose the ability to view that notebook
> Showing /usr/src/nbgallery/app/views/application/_notebook_tile_small.slim where line #21 raised:
> 
> No route matches {:action=>"show", :controller=>"users", :id=>nil} missing required keys: [:id]
> Extracted source (around line #21):
>       span.sr-only
>         ' Owner:
>       a.nounderline.tooltip-narrow title='More by this author' href="#{user_path(notebook.updater)}"    #{notebook.updater.name}
> 
> Trace of template inclusion: app/views/notebooks/_notebook_recommendation_jumbotron.slim, app/views/notebooks/_notebook_info_jumbotron.slim, app/views/notebooks/show.slim
> 
> Rails.root: /usr/src/nbgallery

Revisions table shows NULL for user ID for the initial creator and most of the commits.

mcrutch:
…user has been deleted

image

Closes #359 
(As far as I can tell this addresses all known cases of this problem)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -88,7 +88,9 @@ module ApplicationHelper
   end
 
   def link_to_user(user)
-    if user.org != nil && user.org.length > 0
+    if user == nil || user.id == nil
+      "Unknown"
+    elsif user.org != nil && user.org.length > 0
       link_to(user.name, user, class: "tooltips", title: "#{user.name} (#{user.org})")
     else
       link_to(user.name, user)
--- a/app/views/application/_author_rep_trophy_icon.slim
+++ b/app/views/application/_author_rep_trophy_icon.slim
@@ -1,16 +1,17 @@
 span.hidden aria-hidden="true" #{"["}
-a href="#{summary_user_path(author)}"
-  -if author.author_rep_pct
-    -if author.author_rep_pct >= 90
-      span.glyphicon.glyphicon-equalizer.author-rep.author-rep-gold.tooltips aria-hidden="true" title="Author contribution score: #{author.author_rep_pct.floor}"
-    -elsif author.author_rep_pct >= 75
-      span.glyphicon.glyphicon-equalizer.author-rep.author-rep-silver.tooltips aria-hidden="true" title="Author contribution score: #{author.author_rep_pct.floor}"
-    -elsif author.author_rep_pct >= 50
-      span.glyphicon.glyphicon-equalizer.author-rep.author-rep-bronze.tooltips aria-hidden="true" title="Author contribution score: #{author.author_rep_pct.floor}"
+-if defined? author and defined? author.id
+  a href="#{summary_user_path(author)}"
+    -if author.author_rep_pct
+      -if author.author_rep_pct >= 90
+        span.glyphicon.glyphicon-equalizer.author-rep.author-rep-gold.tooltips aria-hidden="true" title="Author contribution score: #{author.author_rep_pct.floor}"
+      -elsif author.author_rep_pct >= 75
+        span.glyphicon.glyphicon-equalizer.author-rep.author-rep-silver.tooltips aria-hidden="true" title="Author contribution score: #{author.author_rep_pct.floor}"
+      -elsif author.author_rep_pct >= 50
+        span.glyphicon.glyphicon-equalizer.author-rep.author-rep-bronze.tooltips aria-hidden="true" title="Author contribution score: #{author.author_rep_pct.floor}"
+      -else
+        span.glyphicon.glyphicon-equalizer.author-rep.author-rep-white.tooltips aria-hidden="true" title="Author contribution score: #{author.author_rep_pct.floor}"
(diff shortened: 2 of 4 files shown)


manfromjupyter:
1. Just confirming, did we want this functionality for revision history mentioning of an Unknown user? Just think it might be a little hard translation wise for accessibility users if it's "--". Might we also make this say "Unknown"?
revisions - deleted user

2. Need confirmation on this but I think this may have also deleted the notebook metric data for this user so users looking at edit history datatable I see none. Notebook was mass-uploaded so don't think it was ever viewed by said user, nevertheless is showing up in none of the tables that was deleted.

mcrutch:
For #1 - Changed (and changed it to use the link_to_user function for when the user is known)
for 2. The clicks data has to be deleted because user_id is "not null" in the clicks table and is part of a multi-index.  Revisions still exist, but they are set to user_id of NULL.  There are 859 such rows now from the users we have deleted.

manfromjupyter:
I have confirmed that works beautifully. Thank you. I did a little more diving and I found a couple more issues I didn't think to check for earlier.

1. When deleting a user who currently the OWNER of a notebook (not creator or last editor), the notebook gets deleted still. Confirmed via database. I think expected due to Ruby on Rails but don't think we want that.
Former Link: {My URL}/notebooks/81

2. A Notebook deletion does not delete all comments associated with it from the database (unrelated issue). Will file another bug if you agree this is a bug. Have confirmed though that a notebook deleted deletes all reviews and change requests associated with it though.
Former Review Link: {My URL}/reviews/63
Former Change Request Link: {My URL}/change_requests/41

3. Delete a user who proposed a review for a page will still have their name exposed in the comment of the review because of how we do it. Ex. "Submitted by John Smith: 'Needs review (for test)'". The Review process isn't finalized and can easily be changed, just making a note of it here.

4. Deleting a user who was the one who submitted a change request deletes the change request. This to me feels like Ruby on Rails would make it happen but we'd prefer it didn't.
Former Change Request Link: {My URL}/change_requests/42

5. Deleting a user that claimed a review, removes them and render just fine, which is awesome, but the UI of the review does not allow any other user to claim or unclaim (even as an admin). Need to do a check if reviewer_id = Null and status = claimed and if so give option to unclaim. Right now current workaround would just be for an admin to delete the review and then remake it.
 Link: {My URL}/reviews/65
no way to unclaim review

6. Deleting a user who placed a comment on a notebook causes a 500 error to occur due to a nilClass referring a variable. Just needs a null check.
Link: {My URL}/notebooks/80
deleting user who placed comment breaks notebook

7. Deleting a user that submitted a deprecation banner causes the Notebook to 404. Still shows up in database, just won't render.
Link: {My URL}/notebooks/79

jeffyjefflabs:
>     1. When deleting a user who currently the OWNER of a notebook (not creator or last editor), the notebook gets deleted still. Confirmed via database. I think expected due to Ruby on Rails but don't think we want that.

For what it's worth, this was by design.  If you delete your account from, say, a photo sharing site, I think most people would expect their photos to no longer be accessible.  But I can see arguments for institutional knowledge here so I'm not going to object too strongly if you change it.

mcrutch:
BLUF:
1 - Intended, do not change
2 - Separate Bug
3 - That's ok with me so long as that record isn't also tied to their user id as well which will cause render stupidness
4 - Intended If the change request hasn't been merged by the time you delete the user, oh well
5 - Reviews...you delete the user, their stuff goes away.  Users can't even delete themselves so....if an admin is deleting a user, treat it as a purge (Unless I am missing a spot a user can delete themselves, but I can't find it)
6 - Separate Bug...not that we use comments much anyway
7 - Separate Bug to sort out
8 - Bug

Longer version
Delete User cascading to delete Notebook/un merged Change Request/Review owned by them. I prefer it that way then trying to figure out how to deal with so many zombie notebooks/reviews/change requests.  Zombie revisions, fine, because that lets you change the owner and then delete the user if you really want them gone. But I don't want unowned notebooks that are harder to discover/fix lying around. Think about the scenario that a really bad notebook is found later on and now you have no one who "owns it" and no traceability to who did it.  If it has a new "owner" at least you have someone you can talk to about it, but if it has no owner? That makes your entire code base if else statements all over the place.

manfromjupyter:
Sounds good.
Will file the separate bugs. Worry not about that.
I can get behind that for the Change Requests if they don't get accepted in time. 
Might we change he hover text (tooltip) for the trash can icon on users on  the /users page to say "Delete and Purge User Data"? Would let people know the product is GDPR compliant lol

mcrutch:
Where did we stand on this?  Merge away and address the other things in related tickets?

manfromjupyter:
@mcrutch yes, this is good to go in my opinion. This has already addressed a lot we didnt cover and leaves nothing new broken. The subtickets of #360, #362, #363, and #364 will be addressed later. :)

PR #366, opened 2020-11-30
Ensure that the font-size stops growing at a known point rather than …

### LINKED ISSUE #307: Group Header is malformed for extremely wide screens
image
vs
image

mcrutch:
…growing continuously even though the container stops growing

Closes #307 

57px is ~ the size of 3vw at 1920 resolution which is standard for a "1080" monitor that is common these days, but prevents things from going horribly wrong on a 2560x1440 monitor


manfromjupyter (approved):
Code looks good and confirmed it looks good at all widths on Users, Groups, Languages, and Tags pages.

PR #368, opened 2020-12-01
Cascade deleting comments

### LINKED ISSUE #362: Notebook deletion does not delete comments on it from database
A Notebook deletion does not delete all comments associated with it from the database.

### LINKED ISSUE #363: Deleting User Breaks Notebook if they Left a Comment
Deleting a user who placed a comment on a notebook causes a 500 error to occur due to a nilClass referring a variable. Just needs a null check.

deleting user who placed comment breaks notebook

mcrutch:
Delete Comments made by a user when they are deleted
Delete Comments made on a notebook when it is deleted
Closes #363
Closes #362


manfromjupyter (approved):
Approved. Confirmed both use-cases work correctly now without error.

Issue #353, opened 2020-11-16
Subscription Page 404ing when notebook, tag, or group record deleted


manfromjupyter:
Notebook, tag, and group deletions needs to do a check every time to see if a subscription element exists for that particular element and delete them if so so the Subscription page will still load. Subscription email is probably also broken.

Tag deletion may need to be covered as part of the nightly computation before the subscription email. Pretty much do a check for all tag subscriptions that exist and make sure at least 1 still exists or else delete from subscription page (if we have no efficient way to do it on deletion).

mcrutch:
Related #360.
For Notebooks and Groups, might have to see if enough information is available to do a custom destroy on tags for cascading.

mcrutch:
Possible solution for the cascading destroy - 
Set up a "before_destory" on Notebooks and Groups that delete any subscriptions to them:
https://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html

PR #369, opened 2020-12-01
Cascade delete subscription

mcrutch:
MERGE AFTER #368 
Partially addresses #353 
Tags are still a problem...Won't address tags until #360 is done.


manfromjupyter (approved):
Confirmed that deleting Users and Notebooks works like a charm and displayed correctly on Subscription page. There doesn't seem to be a way to delete groups in our UI and deleting it via the database will not trigger the ruby function so just going to assume it also words; it looks right at the very least.

PR #371, opened 2020-12-02
Sets Creator and or Updator to Unknown in solr if the user has been deleted

### LINKED ISSUE #370: Indexing fails if there is a Null notebook creator

mcrutch:
Closes #370


manfromjupyter (approved):
Approved. Confirmed that when deleting Owner, the Notebook is deleted, but when creator or updater is deleted it just replaces them with "Unknown"s on both notebook and notebook listing views.

PR #374, opened 2020-12-02
Add "Upload" button to mobile menu

### LINKED ISSUE #244: Upload button lost when on mobile views
It seems we forgot to add the upload functionality/link to the mobile menu dropdown at the smallest screensizes. It seems like the upload modal will still work at that size so no reason to have a lessened mobile experience.

mcrutch:
Closes #244

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -327,7 +327,7 @@ body.page-admin .navbar-wrapper {
             display: none;
         }
     }
-    @media all and (max-width: 800px) {
+    @media all and (max-width: 500px) {
         #uploadModalButton {
             display: none;
         }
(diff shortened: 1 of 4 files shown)


manfromjupyter:
Your change does satisfy the ticket requirements but I want to change it to more closely match the desktop experience. Think Upload button should be at the top. Also there is a gap inbetween 992px and 800px that both the More Actions icon and Notebook both exist but have the same content. Have a current working branch of these fixes, but just dont think I'm going to finish testing and that before a minute ago lol

manfromjupyter:
Above 920px you see all buttons and things

<920px you lose the Big Notebook dropdown (replaced with the hamburger menu) and the Launch Tour (which exists under the Help icon).

<800px the custom buttons disappear and show up in the More hamburger menu.

<500px the upload button disappears and moves to the More dropdown as a link at the very top. That's what is was before. Original intention of ticket was just to keep it as a button as long as possible (disappears when going from 300% to 400% zoom)

Works with zooming as well (in same order).

In addition removed duplicate ID of Upload button. We don't do it for the Launch tour and still works so why risk the potential HTML or JS violation with or without stylesheet being rendered. Can roll it back if we want to roll it back though.

manfromjupyter (approved):

Issue #352, opened 2020-11-12
Replace User Summary datepicker images


manfromjupyter:
For easier theme customizing, can't be using images, should replace with a background-color or images stored locally that look better with our themes

mcrutch:
Fun fact, they don't even appear in production mode.  I now understand WHY...but swaping it out means replacing the entire icon sheet or doing a lot of serious style overrides.  For now, I am going to force them to appear at least

mcrutch:
FYI -  In the running container, I did this - 

```/usr/src/nbgallery/public/assets# cp -R ../../vendor/assets/stylesheets/images ./```

mcrutch:
Thought....since we already use glyphicons everywhere...bootstrap datepicker?  Since that uses glyphicons rather than images.

PR #382, opened 2020-12-04
Change to bootstrap-datepicker for styling reasons

mcrutch:
This datepicker uses glyphicons and can more easily be styled

Possible solution to #352 by using a different, non-image based, datepicker

The max-width is because the datepicker seems to build it's size as a factor of the input field width and the input field for the warnings page is massively long.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,7 @@ gem 'sdoc', group: :doc
 
 gem 'acts_as_votable'
 gem 'bootstrap-sass'
+gem 'bootstrap-datepicker-rails'
 gem 'browser'
 gem 'chartkick'
 gem 'commontator'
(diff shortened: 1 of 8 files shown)


manfromjupyter (commented):
Works and satisfies our accessibility requirements but with your permission I'd like to few lines of CSS to clean it up a bit

some fixes for bootstrap datepicker

mcrutch:
By all means.  hack away at the CSS!

manfromjupyter:
So I'm happy with it for this branch. Screenshot before for normal and low vision users. Only reservation is the modal supports complex keyboard commands to help select an ideal date (Shift + Left/Right arrow to select current date but different month and CTRL + Alt + Right/Left Arrow to change current by year) but it doesn't let the screenreader users get into it easily, doesn't tell them how to get the popup up and what the month or date is when on a tbody cell. Can't use arrows keys to get up to read them either. These issues existed before but I guess I remember it differently so will file separate issue but just noted it here.

datepicker-edits
lowvision-datepicker-edits

manfromjupyter (approved):

Issue #387, opened 2020-12-14
On home page, error messages within modals cause a padding issue


manfromjupyter:
![home page modal errors padding issue](https://user-images.githubusercontent.com/51969207/102106309-66c36100-3dfe-11eb-86ff-458a1d40ea1a.PNG)

manfromjupyter:
Will do this one soon because it's annoying me. Another photo:

![home page modal errors padding](https://user-images.githubusercontent.com/51969207/105093974-8db22a00-5a71-11eb-90bd-7c5ac3265c29.PNG)

PR #388, opened 2020-12-14
Old title character parsing

### LINKED ISSUE #386: For legacy and non-edited content since then, on update, change title to use full width special characters if character is illegal
Long time ago we made `:`, `/`, and `\` characters illegal for new notebook updates to match that of Jupyter and prevent issues sharing back down. Now Change Request, Upload New Versions, and perhaps other things fail when updating notebook version until the title is edited. Fix this atuomatically by changing `:` to `：` or `︓` and perhaps for the other two leave a message lettings users know if they save their notebook as they are updating that the title will be edited and `/` and `\` characters will be replaces with spaces.

manfromjupyter:
Closes #386 and #387 

Partially for legacy systems that have old notebooks with title with characters we have since deemed illegal so it better communicates with Jupyter. Whenever a user uploads a new version or accepts a change request any of the illegal characters :/\ will automatically switch to full width characters, with exception of the colon which is just a different Jupyter-friendly variant that barely looks different than the normal colon. 

In addition fixed an issue where padding was off for errors on home page and edited current error message for when a new upload has illegal characters to include the approved icons within the error message so they could copy/paste to save time.

## Changed Error Messages
upload bad title error
title edit error

## When Fixed (no the prettiest but it at least lets them still preform the notebook update)
full width characters

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -2221,12 +2221,12 @@ body.page-notebook-search {
         z-index: 1000;
     }
     .external-gallery .run_jupyter {
-      display:none;
+        display: none;
     }
 }
 
-body.page-home .alert,
-body.page-notebook-search .alert {
+body.page-home #main .alert,
+body.page-notebook-search #main .alert {
     margin: .75em 0 -.5em;
 }
 body.page-notebook-search h1 {
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -586,6 +586,21 @@ class ApplicationController < ActionController::Base
     )
   end
 
+  def notebook_title_character_cleanse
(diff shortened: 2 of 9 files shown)


mcrutch:
Can you walk me through the workflow on this one when you have a chance.  Sounds like you are saying you fix it automatically, but then it also seems like the error is still there....maybe I am missing something.

manfromjupyter:
@mcrutch, So if you have a title with a : / or \ (which you can fake by changing title via the database), if you then upload new notebook version or have a change request, this title will automatically get fixed above in the last screenshot.

If you instead try to change title while having those invalid characters or injecting them, you get the better error message telling you you have to use the full width chars and includes them. If uploading a new notebook you get the same. The only time it will automatically fix will be on version changes and change requests because these do not let you change title and these notebooks are otherwise unable to do these changes (until title is fixed - and frankly should be fixed anyway because of it's running/saving in Jupyter).

PR #392, opened 2020-12-15
Landmarks: Added numerous ARIA landmarks for screenreaders as well as proper semantic html fixes

### LINKED ISSUE #365: More Accessibility Landmarks
Bit of a lift, but need to do eventually. To allow screenreaders to jump around the page with a touch of their keyboard, the following landmarks needs to be added/changed.

- Need to reformat header to include the tag="header" with the site warning banner put at the top of it. Then set role="navigation" to all of the nav elements (buttons and dropdowns).
- Add navigation to some other elements with aria labels to differentiate between them such as the home page notebook tabs and notebook actions.
- Some others

Already did main and searches, and no complementary, application, or contentinfo roles because we don't have sidebars, generated keyboard, or a footer.

manfromjupyter:
Closes #365 

Made proper header, added navs, tablist for home page, and made notebook region so user knows to read it as read-only (no real buttons to click on). Tabbing by links, buttons will vary so feel like need to keep it consistent so they know.

Notebook region may need to be a type of article instead so the heading violations (two H1s side by side) are not seen a problem but not sure article really applies so just made it a generic region section with the label "Notebook content"

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -246,18 +246,18 @@ a.skip {
         top: 0;
     }
 }
-.navbar-wrapper {
+header {
     background: $site-header-background-color;
     border-bottom: 1px solid $tertiary-background;
     box-shadow: 0 -2px 8px $tertiary-background;
     margin-bottom: 1em;
     padding: 5px 0;
 }
-body.page-home .navbar-wrapper,
-body.page-notebook-search .navbar-wrapper {
+body.page-home header,
+body.page-notebook-search header {
     margin-bottom: 0;
 }
-body.page-admin .navbar-wrapper {
+body.page-admin header {
     border-bottom: 1px solid $primary-border-color;
     box-shadow: 0 -1px 8px $tertiary-background;
 }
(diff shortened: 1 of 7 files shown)


mcrutch:
Just wondering why you used "header" here rather than role="banner" because the latter is explicit while the former is dependent on where it is in the DOM on whether it gets the equivalent to role="banner" (unless I mis-read the Aria guidelines).

manfromjupyter:
I will confirm in a bit with jumping from region to region, but based on what I read, the role="banner" is only used when a header tag does not exist. Same as role="contentinfo" when tag of footer is not found. Because only one header and footer can be used on a page, for it to be proper HTML, I think it's legit, but will check in a bit and get back to you.

manfromjupyter:
@mcrutch Tested it all on JAWS and confirmed, header tag acts as a region that can be jumped to at any time (doesn't need the banner role and the footer same thing for sitecontentinfo). Also took the time to fix 1 or 2 other aria labels that were just annoying me/something needing fixing. Realized the logo, like 10 fixes later still sounded super wrong. Screenreader was verbalizing it as "n'gallum link" instead of "NBGallery Home link". Seems it trimmed the leading or trailing spaces for multiple elements under same parent element. We still need them elsewhere in the product for the rendering without a stylesheet, but anyway that is why the logo code behind the scenes may look super weird in this fix. It's 100% perfect now for screenreaders, non-screenreader, and decent for stylesheet-free peeps.

PR #397, opened 2020-12-18
Absolute url tech debt cleanup

### LINKED ISSUE #291: Remove all absolute paths from .slim files
There are instances of hardcoded directory paths (such as "/notebooks" on the search form).  These need to be relative OR use rails values to base them on the application root

manfromjupyter:
Fixed up all absolute URLs that can be changed successfully in the views and controller sections. Tested all with exception of a few that were hard to reproduce. Closes #291 

Need to confirm that Running in Jupyter works. I couldn't log into the specific site to test it (for whatever reason).

### DIFF
--- a/app/views/application/_notebook_run.slim
+++ b/app/views/application/_notebook_run.slim
@@ -1,5 +1,5 @@
 -if @user.member?
   div.ribbon-wrapper
     div.github-fork-ribbon
-      a.run_jupyter href='/notebook/#{ nb.uuid }' Run in Jupyter
-      input type='hidden' value="#{(defined?(ref) && !ref.blank? ? ('?ref=' + ref) : '')}"
+      a.run_jupyter href="#{uuid_notebook_path(nb.uuid)}" Run in Jupyter
+      input type="hidden" value="#{(defined?(ref) && !ref.blank? ? ('?ref=' + ref) : '')}"
--- a/app/views/application/_notebook_run.slim
+++ b/app/views/application/_notebook_run.slim
@@ -1,5 +1,5 @@
 -if @user.member?
   div.ribbon-wrapper
     div.github-fork-ribbon
-      a.run_jupyter href="#{uuid_notebook_path(nb.uuid)}" Run in Jupyter
+      a.run_jupyter href="#{notebook_path(nb.uuid)}" Run in Jupyter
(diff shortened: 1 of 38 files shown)


manfromjupyter:
@mcrutch Fixed merge conflicts. Is good to go now on this one

app/views/application/_notebook_run.slim
```diff
@@ -1,5 +1,5 @@
 -if @user.member?
   div.ribbon-wrapper
     div.github-fork-ribbon
-      a.run_jupyter href='/notebook/#{ nb.uuid }' Run in Jupyter
-      input type='hidden' value="#{(defined?(ref) && !ref.blank? ? ('?ref=' + ref) : '')}"
+      a.run_jupyter href="#{uuid_notebook_path(nb.uuid)}" Run in Jupyter
```
mcrutch, line 4: This causes things to break because it has a trailing /uuid on the URL so the code in run_in_jupyter.erb fails when it tries to pull the actual UUID out of the URL which then causes the notebook to not be found.
Old format of this link: https://<domain>/notebook/<uuid>
This code produces: https://<domain>/notebook/<uuid>/uuid
The code that it breaks is:
https://github.com/nbgallery/nbgallery/blob/38c8c48ed2040ee120e16be33011242273c717a4/app/assets/javascripts/run_in_jupyter.js.erb#L17

mcrutch (changes requested):

app/views/application/_notebook_run.slim
```diff
@@ -1,5 +1,5 @@
 -if @user.member?
   div.ribbon-wrapper
     div.github-fork-ribbon
-      a.run_jupyter href="#{uuid_notebook_path(nb.uuid)}" Run in Jupyter
+      a.run_jupyter href="#{notebook_path(nb.uuid)}" Run in Jupyter
```
manfromjupyter, line 4: @mcrutch, This should do it. Good catch. Thought I tested but I suppose not. I lost all of my environments when I accidentally deleted myself a few tickets ago.

mcrutch (changes requested):
Found more problems.  Please correct and do a complete retest to verify.

manfromjupyter:
@mcrutch oh woops, it appears those tag mistakes were from copy/pasting one that worked. That as well as a home page issue I found are fixed now.  Made sure all 144 lines and however many links and forms with href of those are all showing up correctly now. This one should be good to go now

Issue #399, opened 2020-12-21
Huge Notebooks are sluggish to view


manfromjupyter:
Fixing this could also fix #15

PR #400, opened 2020-12-21
Added pagination of code cells on notebooks to help load ridiculously large notebooks easier.

### LINKED ISSUE #15: Notebook size limit
Add a size limit on notebook uploads and/or the size of notebooks rendered.

manfromjupyter:
Closes #15, #399, and #401 

Added pagination after 20 notebook cells (code cells, text cells, etc). If this needs to be changed or moved up, it's an easy fix, but for testing given that the size of these 20 can vary use this as a baseline:

With breaking in 20 cell increments:
{SITE}/notebooks/86-recursion-examples is the minimum in terms of scrolling that would be required.
{SITE}/notebooks/148-visualization-earthquakes?page=4  is the maximum in terms of scrolling that I could find with these changes. Both load super easily.

Relative minimum:
soft min scroll bar

Relative maximum:
soft max scroll bar

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -1877,14 +1877,8 @@ p.description {
     position: relative;
 }
 #viewAsFullpage {
-    display: block;
-    margin-top: -1em;
-    @media all and (min-width: 800px) {
-        bottom: 1.5em;
-        display: inline-block;
-        margin-left: 1em;
-        position: absolute;
-    }
+    display: inline-block;
+    margin-top: -.5em;
 }
 
 /* ====================================== */
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -3292,7 +3292,9 @@ body.dark-theme {
             filter: invert(100%);
         }
     }
-    .sparkline {
+    .sparkline,
+    #viewAsFullPage,
+    #notebookCellPagination li.active span {
         filter: invert(100%);
     }
     #changeReqView {
@@ -3525,7 +3527,8 @@ body.ultra-dark-theme {
     &.path-video video,
     #navbarSearchBar .info-button,
     #expandableSearchDropdown,
-    .add-icon {
+    .add-icon,
+    #viewAsFullPage {
         filter: invert(100%);
     }
     .modal-header {
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -1872,6 +1872,21 @@ p.description {
     color: #fff;
 }
 
+/* === Notebook Pagination === */
+#notebookCellPagination {
+    position: relative;
+}
+#viewAsFullpage {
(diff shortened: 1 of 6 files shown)


mcrutch:
There must be a way to chose to view the full notebook even if pagination is the default view.

manfromjupyter:
@mcrutch, do you think this will be better as an NBGallery User Preference (to be a blanket paginate/non-pagination mode) or some UI element within a given notebook page that makes just that individual page load instead?

mcrutch:
I think paginate by default with a link in the pagination area to expand to single page. Saves processing time/page load times most of the time when people just want to go and click Run, but makes it easy to toggle to full view.

mcrutch:
OR....infinite scroll would be awesome, but might be harder from an accessibility point of view

manfromjupyter:
@mcrutch, how does this sound/look. User can manually include `?view=full` to their notebook to reload the page without pagination (as we see now). Elsewise has pagination and at the bottom next to the pagination is a button link to render the entire thing on one page (current functionality). Is that about what you were thinking and if so, is there anything you want changed? Param name, button title, etc

view full - low pagination
view full - high pagination

Low Vision/Mobile support
low vision support - high pagination

mcrutch:
Thoughts on making the button always on the next line and say "View Full Notebook" instead of just "View Full"

manfromjupyter:
@mcrutch I think the thought process behind the placement of this was to make it clear that View Full was an alternative to pagination and since related should be inlined, sort of like the radio buttons being inlined. Your solution would be easier for low vision and would be more clearly understood what it's purpose is though. I am open to either.

mcrutch:
Yeah, I just feel it flows better especially if a super long notebook ends up with an ... in the pagination because someone MIGHT think it is to view the full list of pages? I dunno, just feels like, since it is an either/or, rather it be a little more distinct from the pagination.

manfromjupyter:
Works for me. Will make the change. Currently wrapped in getting some JupyterLab testing done for some people's tickets so will just make sure to have it all gucci for you on Tuesday or whenever

manfromjupyter:
@mcrutch, it's new-lined and beautiful now. Good to go from my end.

view full notebook button

mcrutch:
My only nitpick (and it is consistent with other stuff, just a little weird) is that in the normal theme, pagination and the button match.  In Ultra dark (and probably others), it looks weird because the styles clash.  I say it is consistent with other stuff, because it is just using the button theme used above on the page, but since the elements are related in purpose, should they instead match in color no matter the theme?

image

manfromjupyter:
Oh you're absolutely right. I completely forgot about themes and should have checked that. Will do the merge conflicts and then resolve this.

manfromjupyter:
Fixed and ready for you. Also appears the full notebook page was sometimes or always appearing under super small notebooks so added a condition to prevent that (triple checked it and it's good).

dark view full
ultra dark view full
regular theme full

PR #402, opened 2020-12-21
Code cell changes

manfromjupyter:
Made it so horizontal scroll bars no longer appear:
code cells with the changes

My attempt at editing the markdown syntax highlighter/markdown plugin to make it look like the current code block but with line numbers and the above issue of wrapped lines having same text indent applied to it was not successful. Best I could do with just a CSS hack for the record was as follows. Changes were not merged for obvious reasons.
line numbers with css hacks purely

Will address these line number and other changes later, but for now leaving the code on here just to wrap them. If we want to turn the horizontal scrollbar functionality as something users can enable in their NBGallery preferences, just let me know, I think we need this out of he box though, although would be a lot better with line numbers or that margin fix.

Another thing to note, different version of a couple different gems will add a `<code>` tag between the pre.highlight and spans. I think this is needed eventually for screenreaders so users know a block is code, but not a requirement to date nor do I know if the experience would be different anyway.

### DIFF
--- a/app/assets/stylesheets/formatters.scss
+++ b/app/assets/stylesheets/formatters.scss
@@ -2,6 +2,8 @@
 /* generated using Pygments.css('.highlight') */
 .highlight {
     background: #f8f8f8;
+    white-space: pre-wrap;
+    word-break: break-all;
     .hll { background-color: #ffc }
     .c { color: #3e7c7c; font-style: italic } /* Comment */
     .err { border: 1px solid #f00 } /* Error */


mcrutch:
So, is this intended to be merged?  The comments above seem to imply you want to do more work on it, so I don't want to merge it if you still want to do other things with it.

manfromjupyter:
Oh hah, sorry. I see it was just a vomit of ideas. Partly I feel disappointed I couldn't get line numbers working; tried so many different combinations and things. Partly I think I need more time to try to get line numbers or hard indents working but I mostly think I just need buy in from what you think we should do. Should we:
1.) Push this as is which is an accessibility fix for low vision users and perhaps also general UX for particular cells with long lines because it will be easier/quicker to read
2.) Make this or the current functionality default functionality and add it as a NBGallery User Preference
3.) Wait to merge this until we get line numbers or indents working and perhaps look at new syntax highlighers?

manfromjupyter:
Definitely can get behind making line numbers off by default and making it a NBGallery User Pref. As a side note, just wanted to let you know that with CSS, I can make the line numbers unselectable and uncopyable

mcrutch:
So, leave this as WIP until you can do numbers, or merge this and adding numbers will come later?

manfromjupyter:
Will look into numbering again.. Give it one more go and then come back to you

manfromjupyter:
Spent another 3 hours on this trying to get kramdown, HTML Pipeline, rouge, or redcarpet working in terms of displaying line numbers or better indents, but couldn't get it to working despite it supposed to be working. May just be our rails version or our implementation but feel like I completely redid it and still wasn't working. Looked at other plugins and gems that work with Ruby and unfortunately there are very few options and we are already using the good ones.

Would like to release current code for the accessibility fix. As per the line numbers and/or spacing, we just can't do it best I have tried with our current version and gems.

PR #417, opened 2021-01-14
Enable option to enable private notebooks in notebook listings when user is admin

### LINKED ISSUE #339: Show private notebooks in Group, User, and search listings for Admins
Currently admins can see private notebooks in reviews and change requests that don't belong to them, but when looking up a notebook by name, or transferring ownership of notebooks from a group or user to another, etc admins need to be able to see private notebooks in said listings.

For searches can enable it with a checkbox like deprecated?

Current workaround is for the admin to go to each notebook directly via url after returning a list of ids in a MYSQL database call.

manfromjupyter:
Will close #339

Having a dev box issue so pushing what I have. Will finish the rest of it in a bit.

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -546,9 +546,10 @@ $(document).ready(function(){
   }).attr('checked', $('#showPrivateNotebooks').val()=="true");*/
 
   $('#filterPrivateNotebooksForm #showPrivateNotebooks').change(function(){
-    if $('#filterDeprecatedForm').prop("checked") {
-      
-    }
+    var params = $('#filterPrivateNotebooksForm').serialize();
+    var url = window.location.href.split("?")[0]
+    window.location = url + "?" + params;
+    //if ($('#showPrivateNotebooks').prop("checked")) {
   });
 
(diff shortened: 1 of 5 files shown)


manfromjupyter:
@mcrutch I think this one is good to go. There is an existing bug where for low vision users the search table is floating off the side of the page and this makes it worse but figured it should be done in another ticket to not delay this.

Tested it and it works as expected with or without deprecated also enabled and sorting also works no problem. Only other thing that may seem off is enabling deprecated and privates and then adding a filter resets the deprecated and privates filter to default settings. I feel like this is expected behavior, but just wanted to bring it up just in case.

The pieces that will get fixed in another ticket if you concur:
new additions to search table  kind of
400% mag on search page

mcrutch:
Good catch on the show_depreacted vs admin, I will flag that in a ticket how I want it.

Changing search clearing checkboxes is fine, although I think we should allow it to be done as a keyword in the search field going forward.  The reason: If the only notebook that matches your search criteria requires admin, or is deprecated, you are never presented the checkboxes because no results are displayed.  I will document this in a ticket too.  

View issue: Yep address as appropriate on another ticket.

👍

PR #418, opened 2021-01-14
Provide Feedback fix: Made it so fields are hidden when they should be

### LINKED ISSUE #316: Providing Feedback "Broken?" field does not disappear once toggling back
When on Provide Feedback modal/form, if you select "yes" you Did run the notebook and then toggle back and forth between the "Did it work?" field from "Yes" to "No", you will see the "Broken?" field that pops up doesn't disappear.

manfromjupyter:
Closes #316

### DIFF
--- a/app/views/notebook_mailer/feedback.html.slim
+++ b/app/views/notebook_mailer/feedback.html.slim
@@ -4,21 +4,28 @@ h2
 br
 p
   a href="#{@url + user_path(@submitter)}"
-    | #{@submitter.name} (#{@submitter.org})
+    | #{@submitter.name}
+    -if @submitter.org != nil && @submitter.org.strip != ""
+      |  (#{@submitter.org})
   |  provided the following feedback:
   i
     |  "
     ==@feedback.general_feedback
     | "
 br
--if @feedback.worked == true
-  p User additionally reports the notebook ran successfully.
--if @feedback.worked == false
-  p User additionally reports notebook is currently broken as it did not run successfully.
-  -unless @feedback.broken_feedback.blank?
-    p
-      ' The user provided the following feedback about what went wrong for them:
-      i
-        |  "
-        ==@feedback.broken_feedback
-        | "
+-if @feedback.ran == nil
+  p User opted out of submitting whether or not they ran the notebook.
+-elsif @feedback.ran == false
+  p User reports they did not run the notebook.
+-else
+  p
+  ' User reports they ran the notebook
+  -if @feedback.worked == nil
+    | but opted out of submitting whether it worked or not.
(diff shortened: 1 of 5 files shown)


mcrutch:
HTML and plain text emails are inconsistent:
HTML:
"xxxx provided the following feedback: "This is feedback""
Plain Text:
"xxxx provided the following feedback: This is feedback Additionally, the user indicated that the notebook was run and it was not successful. The user provided the following feedback about what went wrong for them: I am going to say it worked, but leave this filled out"

Workflow; Click Ran it, click it didn't work, fill out the comment for not working, then change the checkbox to it did work.

mcrutch (changes requested):
Server-side code is inconsistent with how it handles the checkbox being toggled back to Yes it worked after filling in the comment between plain text and HTML emails

manfromjupyter:
@mcrutch, I'm on it. Ill be sure to fix those emails

manfromjupyter:
Made them consistent and while was at it improved the verbiage, spacing, etc. Tested for when notebook was not run, run and no mention of broken, run and labeled broken, and run and labeled as not broken. Good to go on my end

app/views/notebook_mailer/feedback.html.slim
```diff
@@ -11,9 +11,10 @@ p
     ==@feedback.general_feedback
     | "
 br
-- if @feedback.worked == false
-  p
-    | Additionally, the user indicated that the notebook was run and it was not successful.
+-if @feedback.worked == true
```
mcrutch, line 14: You need to check if ran is true as well.
Workflow: Check that you ran it
Check that it didn't work
add feedback for how it didn't work
check that you didn't run it
submit feedback
Email says they reported it didn't work, even though that option was no longer visible on the form.

mcrutch (changes requested):
See comment in code, applies to both text and HTML emails.

mcrutch:
Forgot one other things...check if the org is actually populated before you add the (org) to the email (since you are fixing the views anyway).
    | #{@submitter.name} (#{@submitter.org}) <-- If org is empty, it says `User's Name ()`

manfromjupyter:
@mcrutch Oh smart, will check for org and create another ticket to correct it for all others. Won't take me terribly long to do. As per the running thing, that is wacky. Will fix it though

manfromjupyter:
@mcrutch Should be good to go now. Will fix orgs in other emails later. In terms of the different options when submitting feedback, below will include what both html and raw look like and you can see what all of the other sub options look like in the top 5 emails in the fake email thing. Fixed that submission issue of broken and ran feedback when user switches later they didn't even run it.
html feedback email
text feedback email

PR #432, opened 2021-01-20
Sets Updater and Creator to Unknown in metadata call

### LINKED ISSUE #431: Unable to Run Notebooks when the notebook creator has been deleted
# GET /notebooks/:uuid/metadata
  def metadata
This function doesn't handle null creator/updater

mcrutch:
Closes #431


manfromjupyter (approved):
Golden

PR #435, opened 2021-01-21
Exclude clicks on deprecated notebooks when picking those for review

### LINKED ISSUE #408: Make it so automated nominations of reviews don't occur for deprecated notebooks
Have the automated nomination nightly_computation ignoe deprecated notebooks.

mcrutch:
Closes #408

### DIFF
--- a/app/models/review.rb
+++ b/app/models/review.rb
@@ -88,6 +88,7 @@ class Review < ActiveRecord::Base
     def relevant_clicks(days)
       Click
         .where(action: ['ran notebook', 'executed notebook', 'downloaded notebook'])
+        .where("notebook_id not in (select notebook_id from deprecated_notebooks)")
         .where('updated_at > ?', days.days.ago)
     end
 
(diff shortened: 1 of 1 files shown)


manfromjupyter:
@mcrutch So I deprecated a few popular notebooks and then ran `bundle exec rails runner script/run_job.rb nightly_computation` and your code did work. Those ones did not get chosen but for whatever reason, the Reviews page now looks like this. It's almost like the reviews didn't properly do the check for ones that are already in progress and skip them. I can research into it tomorrow. Might have been due to database changes since my commit messages one too... Idk.

reviews after nightly computation

mcrutch:
Those extra rows are for the "Compliance" review which I turned on during the test, so the Notebook Reviews table doesn't handle those correctly.

manfromjupyter:
@mcrutch Then this is good to go on my end then, unless you wanted me or you to get it working for compliance reviews and then submit just to make sure

mcrutch:
I made that it's own ticket #439. That problem exists with or without this pull request.

PR #436, opened 2021-01-21
Docker dev updates

jeffyjefflabs:
Cleaned up the docker documentation after actually trying it from scratch for the first time.  The old manual setup notes are pretty useless at this point, so I deleted them.  Added some support to set up an environment for doing dev inside docker, with source mounted into the container.


mcrutch:
I haven't forgotten about this, just hasn't been high on my radar lately

PR #437, opened 2021-01-22
When a user is deleted, set any reviews that are "claimed" to "queued…

### LINKED ISSUE #404: Deleting user who claimed a review does not return review to queue
Ignore the circled name at the bottom of the image.. That's not an issue. Thought it was.
claimed reviews when deleting user

mcrutch:
…" but do not clear complete reviews

Closes #404

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,6 +2,7 @@
 class User < ActiveRecord::Base
   before_destroy { |user| Commontator::Comment.where(creator: user.id).destroy_all }
   before_destroy { |user| Subscription.where(sub_type: "user").where(sub_id: user.id).destroy_all }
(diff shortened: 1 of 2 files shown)


manfromjupyter (commented):
Works as expected and doesn't break the reviews page now! Woot woot! Only issue I found is for completed reviews, I think just need to either make it "Completed by Unknown on..." or "Completed on..."

In addition, if review is already completed but reviewer_id is not set, might want to make that also Unknown (because a user can't just unset themself after completing)

completed by unknown needed
completed unknown

mcrutch:
I can look for those views to make those two changes.

manfromjupyter (approved):
Gucci

PR #438, opened 2021-01-22
If a revision was specified, show the updated date for that revision

### LINKED ISSUE #208: Old Revision of Notebook should reflect date when that version was last updated, not the most recent update for the notebook as a whole
Old Revision of Notebook should reflect date when that version was last updated, not the most recent update for the notebook as a whole. This also applies to the user. The most recent editor may be ex. Danny but he never touched that old version, it was John.

Lastly which may be too much of a lift for this ticket but should be considered eventually is that the review status reflects the most recent version. For now should just hide it, but eventually should make it so if both were passed, a verified icon appears for this notebook version.

mcrutch:
Else show the updated date for the notebook
only used if the "updated" code is triggered
closes #208

### DIFF
--- a/app/views/notebooks/_notebook_info_jumbotron.slim
+++ b/app/views/notebooks/_notebook_info_jumbotron.slim
@@ -200,7 +200,10 @@ div.content-container
             ==link_to_user(@notebook.creator)
           -else
             ' Updated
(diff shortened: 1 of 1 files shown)


manfromjupyter (approved):
Oh it's beautiful! Approved (tested and it looks good)

PR #442, opened 2021-01-28
Revert "Commit messages"

mcrutch:
Reverts nbgallery/nbgallery#222


mcrutch:
Reverting because the commit messages code prevents assets from precompiling due to a Javascript bug that javascript can't compile.

PR #445, opened 2021-02-02
WIP: Add Support for Timezones

### LINKED ISSUE #192: Add support for timezones
Timezone dropdown has already been added to User Preferences page, just commented out and has already been added as database column too. All that is needed for this ticket is uncomment that timezone dropdown and then adjust the revealed datetime across the product (which we moved to a single file (_time_ago.slim)) based on the timezone they set.

manfromjupyter:
Closes #192 

In NBGallery Preferences can set timezone where all times when hovered across the rest of the product will now reflect said time  in relation to said timezone (in 24 hour clock). In addition will reveal timezone abbreviation for US and Canada timezones as well/also the UTC+/- number for reference. Could add more but a large number of them have conflicting names (vary country to country).

One problem I see is I use both GMT and UTC. I know they are used interchangeably in a casual environment but 
@mcrutch you have more experience with timezones so wanted to know what one I should make all of them say if we wanted to move to just using one for consistency... UTC or GMT? 

## How to set it
timezones dropdown
timezones

## Negative timezones
date revisions

## Positive and non-abbreviation timezones
positive timezones

## Zero Time Zone (GMT/UTC)
Is just current experience

### DIFF
--- a/app/views/user_preferences/index.slim
+++ b/app/views/user_preferences/index.slim
@@ -80,21 +80,20 @@ div.content-container
           ' Enable Ultimate Accessibility Mode
           strong #{"[BETA]"}*/
 
-    /*h2 Timezone*/
-    /*p Will adjust all times on the site as you see them into what the respective time is for your timezone.*/
-    /*select name="timezone" id="timezone"
+    h2 Timezone
+    p Will adjust all times on the site as you see them into what the respective time is for your timezone.
+    select name="timezone" id="timezone"
       option value="-1200" (GMT -12:00) Eniwetok, Kwajalein
-      option value="-1100" (GMT -11:00) Midway Island, Samoa
-      option value="-1000" (GMT -10:00) Hawaii
-      option value="-950" (GMT -9:30) Taiohae
-      option value="-900" (GMT -9:00) Alaska
-      option value="-800" (GMT -8:00) Pacific Time (US &amp; Canada)
-      option value="-700" (GMT -7:00) Mountain Time (US &amp; Canada)
-      option value="-600" (GMT -6:00) Central Time (US &amp; Canada), Mexico City
-      option value="-500" (GMT -5:00) Eastern Time (US &amp; Canada), Bogota, Lima
-      option value="-450" (GMT -4:30) Caracas
-      option value="-400" (GMT -4:00) Atlantic Time (Canada), Caracas, La Paz
-      option value="-350" (GMT -3:30) Newfoundland
+      option value="-1100" (GMT -11:00/SST) Midway Island, Samoa
+      option value="-1000" (GMT -10:00/HST) Hawaii
+      option value="-930" (GMT -9:30) Taiohae
+      option value="-900" (GMT -9:00/AKST) Alaska
+      option value="-800" (GMT -8:00/PST) Pacific Time (US &amp; Canada)
+      option value="-700" (GMT -7:00/MST) Mountain Time (US &amp; Canada)
+      option value="-600" (GMT -6:00/CST) Central Time (US &amp; Canada), Mexico City
+      option value="-500" (GMT -5:00/EST) Eastern Time (US &amp; Canada), Bogota, Lima
+      option value="-400" (GMT -4:00/AST) Atlantic Time (Canada), Caracas, La Paz
+      option value="-330" (GMT -3:30) Newfoundland
       option value="-300" (GMT -3:00) Brazil, Buenos Aires, Georgetown
       option value="-200" (GMT -2:00) Mid-Atlantic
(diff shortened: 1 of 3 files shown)


mcrutch:
This is not how timezones work.  DST with different time windows between different countries. Some countries and even US states not doing DST.  Timezones are a solved problem in every modern programming language.  Leverage that.

manfromjupyter:
Alrighty, will look into what rails datetime functions, etc can do to do this.

mcrutch:
https://thoughtbot.com/blog/its-about-time-zones <-- here is a start

manfromjupyter:
Super cool. Will make it happen

app/views/user_preferences/index.slim
```diff
@@ -80,66 +80,62 @@ div.content-container
...
-      option value="1000" (GMT +10:00) Eastern Australia, Guam, Vladivostok
-      option value="1050" (GMT +10:30) Lord Howe Island
-      option value="1100" (GMT +11:00) Magadan, Solomon Islands, New Caledonia
-      option value="1150" (GMT +11:30) Norfolk Island
-      option value="1200" (GMT +12:00) Auckland, Wellington, Fiji, Kamchatka
-      option value="1275" (GMT +12:45) Chatham Islands
-      option value="1300" (GMT +13:00) Apia, Nukualofa
-      option value="1400" (GMT +14:00) Line Islands, Tokelau*/
-    /*span
+    h2 Timezone
```
mcrutch, line 83: What about https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/time_zone_options_for_select?

manfromjupyter: So checked it out and although easier and slightly better for low vision users, it is just showing cities, doesn't include day light timezone changes and in my testing of them out, probably 15-20% of them are not working as they should because they are not returning the correct current times from those regions. I don't know how, might be how I'm using it, but it just is.
Rails one liner:
one line rails TimeZone select
Mine:
my implementation

Been following timeanddate.com for accuracy and insight into the unique timezones that Rails supports that don't conform to those around or do/don't practice daylight savings and marking accordingly. Will only take me another hour to be done with it if you like the dropdown thus far or want me to do something else for this (perhaps make less wordy)?

mcrutch: If we don't have a working programmatic source of timezone data for the drop down list, we will not implement this feature. Trying to hand generate the list is going to end up taking a lot of time to do properly to account for all the cases in timezones. Examples from your hand generated list.:
"Indiana (US East)" exists because most of Indiana does not observe Daylight saving time.  
Most of Arizona doesn't do DST (but I don't remember the proper name for this
Neither Bogota nor Lima observe Daylight Savings time, so can't be grouped with US Eastern
And that was just from looking at the US Eastern line, I haven't dug deep on the other times.

Also, from a styling perspective, looking at Windows, Linux and a couple web examples, it is generally also limited to displaying the standard (none DST) offset (UTC-5 for Eastern) not both.  Many things (especially linux based) do use city names as anchors as that can be easier to reference for some people.

manfromjupyter: The Bogota and Lima were left on there from the old list. Fixed it. Indiana time zones are still CST or EST, they just pick the one that is relevant. Yes Arizona doesn't do Daylight savings which is why Rails made them their own endpoint and I made it's own select option (they observe mountain time all year round). We can switch to cities, this just seemed more complete.

I understand you worry about the time suck of doing it by hand, but I did already and got some education on the whole thing. Some of the cities Rails provided didn't work so I used others in the same timezone and those worked. I didn't do one last test yet or lookthrough to see if I mistyped or forgot something and still want to do a little bit more with the ticket, but this is what we're looking at to date (if you are not opposed). Yes timezones change because of global governments but because the full definition of the timezone exists (what the abbreviations are and what it switches to. ex -4 to -5, they will know if it's wrong and they aren't hard to fix should ex. Brazil declare to go back to using daylight savings. If you want me to can it, I can, but it works and I think it's valuable. Certain governments, ex. Myanmar being unstable and potentially changing things shouldn't be a blocker.

top third
middle third
bottom third

mcrutch: Ok, I stand corrected about Indiana (Indiana US East is still listed separately in most software I have looked at for legacy reasons didn't realize they joined everyone else in 2006).  But, back to Bogota and Lima for example. Bogota is in Columbia Standard Time (which isn't on the list, nor is any other -0500 timezone).  Also, since I just did a test between in the OS and in the container and got a (VERY SLIGHTY but still) different timezone definitions, I am not willing to put out code that uses a hard coded list that might not exactly match what the OS lists as available and cause other issues down the road.  If the list is not generated based off of what is actual available on the system executing the code, it will not be merged.

PR #454, opened 2021-02-11
Stop emailing the user who did the sharing saying the notebook was sh…

### LINKED ISSUE #448: Shares a notebook emails the email you already shared it with again
Introduced to the user that reported it to use groups, but still poses a mild annoyance that can maybe addressed.

To reproduce:
1. Share notebook with a user.
2. Add new user to shared notebook and notice that the first user also gets emailed (again).

mcrutch:
…ared with them

closes #448

### DIFF
--- a/app/mailers/notebook_mailer.rb
+++ b/app/mailers/notebook_mailer.rb
@@ -7,7 +7,7 @@ class NotebookMailer < ApplicationMailer
     @owner = owner
     @message = message
     mail(
-      bcc: emails + [owner.email],
+      bcc: emails,
       subject: "NBGallery notebook shared with you"
     )
   end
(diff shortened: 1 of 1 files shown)


manfromjupyter (approved):
Works. Approved.

Sharing with a non_member (a made up user email) doesn't send an email out but I don't think it ever has. Maybe because it accepts only usernames, not emails (and for some instances we have emails as people's usernames). For randos not in the system, they don't have a username to reference. Should ideally throw an error message if we're doing it wrong or strip out all of the code for non-member shares if it can't accept it. Maybe?

mcrutch:
There is a discussion about the non-member stuff in another ticket.

PR #456, opened 2021-02-11
Change show_deprecated from being a 1 to true

### LINKED ISSUE #451: Change show_deprecated param to use true rather than 1
True appears to work fine rather than 1.  use_admin uses true already.  make show_depreacted use true.

mcrutch:
Closes #451


manfromjupyter (approved):
Works and looks good

Issue #453, opened 2021-02-11
Allow show_deprecated and use_admin flags to be set in advanced search form


mcrutch:
Use case: If you are looking for a notebook and the only notebook that matches the search criteria is private or deprecated you won't be presented with the Admin and Deprecated checkboxes because you get the "Sorry no notebooks found :(" screen instead.

At the very least when expanding the advanced search form have them as checkboxes....I don't think they need to be quick-action keywords like user: owner: etc...

mcrutch:
So...at the very least I want to preserve the "use_admin" and "show_depreacted" flags if they just change the search terms, so that is an easy win I am going to push up, but in the long run I would like an easy way to set it (either in the "Sorry no notebooks" message or in the advance search form.

PR #458, opened 2021-02-11
Search preserve flags

mcrutch:
Related to #453 Saves the flags so you can change/refine your search terms without losing that you had toggled the filter.


mcrutch:
Merge after #451

manfromjupyter (approved):
Works splendidly. After the #453 or the other for just when no results get returned, this will be exposed so users can untoggle or enable those options

PR #459, opened 2021-02-11
Throw an error if users try to add a user who doesn't exist to a group

### LINKED ISSUE #423: Groups don't throw errors if adding a new user that doesn't exist

mcrutch:
Closes #423


manfromjupyter (approved):
Looks good and works.

PR #460, opened 2021-02-11
Redesign-Search-Toggles

### LINKED ISSUE #453: Allow show_deprecated and use_admin flags to be set in advanced search form

manfromjupyter:
On search pages now have the private and show deprecated toggles appear under the filters section.

Closes #453 

## With Advanced Search not toggled
search not expanded

## With Advanced Search toggled
advanced search expanded

## Firefox
firefox for search

## Low Vision
400% for low vision on search

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -2062,10 +2062,6 @@ body.page-notebook-search {
         }
     }
 }
-#filterDeprecatedForm,
-#filterPrivateNotebooksForm {
-    margin: 0;
-}
 
 /* === Search tooltip === */
 .search-tooltip {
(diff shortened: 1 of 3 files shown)


mcrutch:
Checkboxes reset when you change the search term which is not the expected behavior.  I might loop back and take a look at this but they should remain in place if you change the search terms and hit enter

manfromjupyter:
@mcrutch Ready again for you. Changing search should no longer reset show_private and show_deprecated. If one or neither are set, they won't be included in the params and not in the new search. Should be gucci

PR #462, opened 2021-02-12
Create the relationship between user and user preferences to cascade …

### LINKED ISSUE #450: (Minor) Delete user preference when user is delete
I apparently didn't make it right or forgot that piece of it (maybe forgot to make user_id a foreign key?). Anyway super minor as the table is already super small and people don't delete users often. This doesn't cause any errors it would just be for a cleaner database. Just need need to delete user_preference row where user_id = said user when a user gets deleted.

mcrutch:
…the deletion

Closes #450

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,6 +16,7 @@ class User < ActiveRecord::Base
          :confirmable, :omniauthable, :timeoutable
   has_one :preference, dependent: :destroy
   has_one :user_summary, dependent: :destroy, autosave: true
+  has_one :user_preference, dependent: :destroy
   has_many :identities, dependent: :destroy
   has_many :environments, dependent: :destroy
   has_many :notebooks, as: :owner, dependent: :destroy, inverse_of: 'owner'


manfromjupyter (commented):
I have no idea why it didn't work, perhaps doesn't have a foreign key, but upon deleting a user with the same user_id as appears in the User Preferences table, the preference didn't delete.

Tried rerunning the server and deleted dummy user 192 as well after updating user preferences row to id of 205... same result
user preferences didnt delete

mcrutch:
Did you restart the container or rails server after checking out this branch?  I tested this and got the expected result before I put in the pull request.

manfromjupyter:
I did yes (which is why I did it twice just to make sure), but will try it again

mcrutch:
Interesting, because when I tested it, it fired off fine.  AT least I am pretty darn certain it did.

manfromjupyter (approved):
Looks like I maybe just forgot to git pull from the `main` branch before checkout out `cascade_user_preferences`. Anyway it works now for me. Confirmed.. approved

Issue #464, opened 2021-02-12
Subscription Bug


mcrutch:
Something involving tags, error message isn't helpful:

'Error: Stopping daily_subscription_email job. Full message: "undefined method 'tag' for #<Tag:ActiveRecord_Relation:xxxxxx>"'

PR #465, opened 2021-02-12
Org in email

### LINKED ISSUE #440: Hide org parentheses in ALL emails if user does not have an org
Changes thought of from #418 

Just use for example:
```
-if @submitter.org != nil && @submitter.org.strip != ""
  ' (#{@submitter.org})
```

mcrutch:
Testing of the daily emails is PARTIALLY blocked by #464 I was able to test several paths, but haven't figured out what path of execution is causing the subscription bug for only my account and not the other test accounts.

Uses .blank? because that accounts for nil, '' and a string of spaces only
Feedback/Change Request emails fully tested
Closes #440


manfromjupyter (approved):
Tested a couple and verified all look good via code. Thank you for doing this one. Sorry I never got around to it. Looks great. Approved.

PR #476, opened 2021-02-26
Allow GET /tags/* in CORS

### LINKED ISSUE #473: Allow GET to /tags/* in CORS config
This is needed to support an extension similar to our autodownload extension.

manfromjupyter:
closes #473 

@jeffyjefflabs and @mcrutch  I don't know to test this so can you let me know if this looks right or if it needs to have a different origin other than (`origins {|_source, _env| true}` that most of the other CORS policies have).


jeffyjefflabs (approved):
Looks right to me.  I don't know a good way to test either, unless you can do it from the browser console somehow?

PR #479, opened 2021-03-02
Allow admins to run a reindex (with useful errors) in the UI

### LINKED ISSUE #477: Allow Notebook and Group reindex from Admin interface in the UI
If a notebook file goes missing or gets corrupted on disk and can't be parsed as a Notebook, the full notebook reindex will fail every time it hits that notebook, but there aren't any useful error messages generated during the nightly cron to find what notebook failed.  Also, sometimes you don't have simple access to the rails command line but need to reindex after an upgrade. 

This will allow an admin to run the reindex in the admin interface for both Groups and Notebooks and if the Notebook.reindex function fails, it will attempt to index each individual notebook and note any notebooks that failed to index.

mcrutch:
Closes #477

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -3230,6 +3230,9 @@ body.page-user_preferences {
     border-color: $primary-border-color;
     margin: 15px 1%;
 }
+body.page-admin-run_reindex h2 {
+    font-size: 200%;
+}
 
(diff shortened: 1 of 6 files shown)


mcrutch:
image

manfromjupyter (approved):
Approved. Looks good and works well. Also made the following design and accessibility changes (I assumed it was okay). 

warning changed
run reindex changes

PR #489, opened 2021-03-09
Database save

### LINKED ISSUE #444: Options to replace local on-disk git repo and notebook storage
Storing the notebooks on disk and the way the git repo operates makes it very difficult to horizontally scale NBG and have a zero-downtime environment.  What can be done to rectify this?  Use this ticket to gather ideas.  

NOTE: This is a heavy lift and may not come to fruition but I want to brainstorm ideas to be able explore a more fault tolerant environment that scales better.

Use a document store to store all of them:
Replace the GitRepo class with something that talks to a document store like MongoDB, or store them in ElasticSearch or something.  Generate UUIDs or use built in facility for UUIDs for commit IDs going forward, make a migration path.

Also have to change how staging and final notebooks are saved in this case, but could be made configurable.

Problems:
Still have to deal with ensuring scheduled jobs only run once even if scaled out, so some engineering would be required there as well.

Why not SOLR?  
Not it's intended purpose

Why not just store them in the database?
Also a possible viable option and makes it easier to ensure consistency avoiding the repo not matching the database problem.

Discuss!

mcrutch:
Please start reviewing this, but do not merge.  So far, everything I have tried has worked with this latest cut of the code, but I am open to suggestions for how to improve it.  Things are complicated by the fact that the NotebookFiles class isn't always used so I had to play it a bit loose with relationships.

Closes #444


config/settings.yml
```diff
@@ -36,6 +36,7 @@ directories:
 # Storage options
 storage:
   track_revisions: false
+  notebook_file_class: true
```
mcrutch, line 39: This should actually be set to false by default, I will change that before the final merge.

app/models/change_request.rb
```diff
@@ -57,7 +67,17 @@ def proposed_notebook
 
   # Set content in file cache
   def proposed_content=(content)
-    File.write(filename, content)
+    if GalleryConfig.storage.database_notebooks
+      notebookFile = NotebookFile.where(save_type: "change_request", uuid: reqid).first
+      if (notebookFile.nil?)
+        notebookFile = NotebookFile.new(save_type: "change_request", uuid: reqid)
+      end
```
jeffyjefflabs, line 74: Could use find_or_initialize_by here

app/models/revision.rb
```diff
@@ -61,9 +64,19 @@ def init
     # Helper for recording a notebook revision
     def notebook_commit(revtype, notebook, user, message)
       return nil unless GalleryConfig.storage.track_revisions
-      commit_id = GitRepo.add_and_commit(notebook, message)
+      if GalleryConfig.storage.database_notebooks
+        commit_id = Digest::SHA1.hexdigest(notebook.content + user.user_name + message + notebook.uuid + DateTime.current.to_s)
+        Rails.logger.debug("Revision Content" + notebook.notebook.to_git_format(notebook.uuid))
+        notebookFile = NotebookFile.new(save_type: "revision", content: notebook.notebook.to_git_format(notebook.uuid), uuid: notebook.uuid)
```
jeffyjefflabs, line 70: It doesn't matter to me either way, but we may not *need* to do the git_format conversion here anymore since it's not actually going into git.  But if it simplifies other code elsewhere, then I have no objection.

docs/database_storage.md
```diff
@@ -0,0 +1,27 @@
...
+appropriate configuration setting to allow database storage.  At this point, the application
+will not work correctly until you migrate the notebooks so you will want to schedule some downtime.
+
+Next, on the command line run `bundle exec rails runner script/migrate_to_db_storage.rb`
+
+The output of this command will tell you if any notebooks were unable to be migrated, this is usually caused by
+a revision or a notebook being missing from the file system. This means content is actually missing
+and migration would not be possible for those specific notebooks or revisions.
+
+**NOTE: These notebooks and revisions are already inaccessible and attempts should be made to purge them from the system**
```
jeffyjefflabs, line 27: Maybe also add a note that reverting from database storage back to on-disk storage isn't supported?  (at least in the sense that we have no plans to develop an "un-migrate" script)

app/models/revision.rb
mcrutch: You know what, I thought there were a bunch of places that did the to/from git content but it turns out it is only the revision.content functions, so I will take a look at this.

Issue #413, opened 2021-01-12
Error handle and enforce requiring emails not usernames for sharing notebooks


manfromjupyter:
When sharing a notebook with a user and adding them by their username and not their email (unless they're the same obviously) causes the share to silently fail and then refreshes the page. Need to be more dummy proof so recommend:
- Change input field's type to that of "email" so the browser disables the submit button until an email is entered
- Update the placeholder text from "Enter usernames" to "Enter emails"
- Enable some kind of error handling if this browser handling of our view fails and the user email was not entered in correctly or whatever.

jeffyjefflabs:
Yeah this was always gross and I don't know what the best strategy is.  On the one hand users don't necessarily know each other's emails since the site should keep those private (except in scenarios where they match, of course), but on the other hand some sort of notification is implied.  Maybe part of the problem is the way we use "share" to mean not only "hey you should check out this notebook" (which only requires an email, regardless of whether the shared-with user has an account) but also "you're now a part-owner" (which only makes sense if the shared-with user *does* have an account, in which case sharing by username makes sense).

manfromjupyter:
Oh I see the predicament lol. Geez

mcrutch:
I think sharing should stay as "username" only allow tokenized values, but allow searching for users by username OR email address (but still only tokenize the user name).  In fact, it should probably act more like group membership than it currently does so it is clear that they aren't RE-sharing with the same people if they go to share again, but instead, keeping them on the list of shared people.

mcrutch:
We will continue to default to usernames, share by email is disabled by default in the pull request I just submitted (since the code to share by email is a completely stub and doesn't send any emails) and an appropriate error message is generated about it being shared by username only if share by email is disabled.

mcrutch:
Decided the dialog as-is is fine and the functionality change of the linked Pull request will count as closing this.

PR #496, opened 2021-03-11
GITHUB-413 - Partial fix

mcrutch:
Since it is stubbed out by default, prevent sharing by email address, make it configurable, and update error messages as appropriate.

Partial fix for #413 - No more silent failures if share by email hasn't been configured.


manfromjupyter (approved):
Prefer error images that aren't formatted like server responses but seems legit and looks good.

prefer errors formatted like sentences

manfromjupyter (approved):
When I share it with your email behind the scenes it neither accepts nor denies it. Doesn't add it or throws an error. Adding someone by username works (and have autocompleted suggestions). In additional adding user who has an email for a username works (as expected). Think only thing off is that first part in how we may want to handle that but approving ( know you said partial fix)

mcrutch:
Cannot replicate sharing it with an email address and it being silent, even using email addresses associated with a user(but not the username)

PR #502, opened 2021-03-15
Make reindexing from the admin page happen in batches

### LINKED ISSUE #498: Admin interface notebook reindex - Batch
For very large datasets the reindex takes too long to do in one process.  "chunk" the reindex sequence and do it with recurring api calls that basically paginate the records.

mcrutch:
closes #498


mcrutch:
@manfromjupyter I welcome any accessibility tweaks that need to happen.  Talk to me about testing it in a way to get errors and to have it actually paginate on our dev site (since our dev site has a whopping 27 notebooks right now after the cleanup).

manfromjupyter:
Still need to review for many many more notebooks but the visual changes are as follows if you dig. The alerts for screenreaders are gucci now but one I have more can verify the ones inbetween started and ending aren't too annoying (if we decide we want them).
loader icon moves down

manfromjupyter:
@mcrutch take a look at those changes and see if they're okay with you. Added enough notebooks for it to do a second listing/page of reindexing. Works successfully and calls it out to screenreaders correctly.

manfromjupyter (approved):

mcrutch:
other than the 

     }
     else {

style making me physically ill, I am ok with this moving forward so you can finish all the necessary merges since I might be out today and tomorrow.

PR #505, opened 2021-03-22
Revisions optional

### LINKED ISSUE #478: Revision saves and commit messages should only be part of workflow when track revisions is enabled
These lines block saving when TRACK_REVISIONS is turned off
https://github.com/nbgallery/nbgallery/blob/b1ec57199b8a4e3ad3efd87244c41bcd06fe2ff8/app/controllers/notebooks_controller.rb#L121
https://github.com/nbgallery/nbgallery/blob/b1ec57199b8a4e3ad3efd87244c41bcd06fe2ff8/app/controllers/notebooks_controller.rb#L151

### LINKED ISSUE #503: Error uploading/updating notebooks: NoMethodError revision.commit_message for nil:NilClass
Hi, 

I am getting errors when uploading or updating notebooks.
The errors are due to revision.commit_message which causes this error: "NoMethodError in NotebooksController#create undefined method `commit_message=' for nil:NilClass"
Not sure if this is due to revision == nil

Thanks

mcrutch:
Merge after #489 
Closes #503 
Closes #478


manfromjupyter:
So it looks good and testing new uploads, version updates, and change requests, and visiting revisions page when I am tracking revisions but didn't try it without because haven't gotten it to work in the past and because I had a question about the /config/initializers/gallery.rb. It looks like this will revert to `true` every time anyway. Is that right?

mcrutch:
You cannot pass ANY environment variable for GALLERY__STORAGE__TRACK_REVISIONS in, because otherwise it is acting like it is always true even if it prints out false, but setting it to false via environment variable is non-standard anyway.  So just remove that environment variable entirely from your docker run statement to test it without track revisions on.

manfromjupyter:
Gotcha gotcha

manfromjupyter (approved):
Works splendidly when no revisions disabled. Upload, new version, change requests, the revision link, and mention of change logs in emails, forms, etc all gone and works well. Can still get to /revisions page on notebooks despite the link being gone by going directly to it but it just says no revisions are available on new notebooks and so that seems legitimate because for notebooks that were added while revisions were added, it doesn't break them. You don't see link to see revisions but can see their previous ones.

PR #506, opened 2021-03-23
Make the "Broken" feedback field disappear if they check that it ran …

### LINKED ISSUE #494: Provide Feedback Fix undid itself it seems. Fix it

mcrutch:
…successfully

Closes #494


manfromjupyter:
So this does fix the issue where the broken field is appearing after changing the did it work field from no to yes, but persists when you change it from if you even ran it from yes to no. I could have sworn I fixed this in https://github.com/nbgallery/nbgallery/pull/418/files but maybe due to a merge conflict it didn't take. So weird. If you don't get around to fixing that part of it, I can do it, just think it's important for the user experience.

Screenshot: (provide feedback still showing broken field when switching to not having run)
provide feedback still showing broken field when switching to not having run

mcrutch:
I think I know what happened.  Shortly after this was merged, I merged the commit-messages code, which had something really broken in it at the time that was a show-stopper and I was trying to deployed to J@H, so I had to roll that back.  Once it was fixed, I recommitted/merged, but somethin got screwed up in that time window.  

We can either roll with this, or you can try applying your same fix as before to a fresh branch off main.

manfromjupyter (approved):
Adding some feedback email additions I added to the previous one as well as the updated controller. I didn't change the JS but although they look different, your works the same so just going to keep it. Other than that, it's all good to go. Tested all 4 types of emails in html and plaintext and both work. And fully tested the form. Entering information in fields I then hide but then submit don't secretly push the information. Everything works as expected

PR #507, opened 2021-03-24
Keep change requests

### LINKED ISSUE #480: Request to Save Accepted, Declined, and Cancelled Change Requests
Assigning to both of you to discuss. Talked to Crutch but he wants to have the conversation with you about it too.

Currently, after 7 days, I estimate, of a change request being approved, declined, or cancelled the dif these point to get deleted which cause the row of the sql table for change_requests to be deleted.

**Pains/What problems this causes:**
1.) For that information older than 7 days, we as admins and fellow content users have no knowledge of who made the change requests, what they comment, who approved/denied it, what the approver/denier/canceller commented, what revision was from a change request (if that matters), and other general knowledge that comes up that people ask about or want to know about, like a user earlier today.
2.) There are two different link locations we have in the product right now that will hit production next time we merge that will return 404's because they link to these that after 7 days will not longer exist. These locations are in the `/notebook/id/revisions` that link to `change_request/id` when version changes due to a change request and also on the /change_request page linking to the same spot so that people can see the timeline of the change request (multiple comments get confusion, especially with the recent addition of the changelog). In addition, the comment bubbles to date alone aren't readable by screenreaders so wanted to make this more user friendly.

In the database it appears the `reqid` column caries the reference to the dif and this is what we are tryin to clean up and why we delete to date (to my knowledge). Anyway, because this is references as a `UNI` key (assume Unique or foreign key), when this gets cleaned up, it takes the entire row with it. Request is to not delete the row and keep the row of data for our records and make available to the users as it's important information. When the request is approved, declined, or cancelled, the dif is not viewable anyway so don't agree we don't need it.

**Solution?**
Not sure if it's this easy but would just entail deleting the revision wrapped up within the change request without deleting the change request metadata, don't show any "resolved" change requests on individual user's `/change_request` pages after 7 days just so it doesn't get cluttered, but for admins on `/change_requests/all` they will all populate and we can add pagination. Because the ID's will still exist, the records will still be accessible to users that care about the timestamps and comments or what got forgotten and we won't have to tell users to just look through their emails to around 2 years ago for it (if it's even about them).

mcrutch:
Closes #480 
Merge after #489


jeffyjefflabs (approved):
Looks fine to me.  I was looking for other places that `proposed_(notebook|content)` get called and the only one that possibly concerned me was in [compare.slim](https://github.com/nbgallery/nbgallery/blob/9c8e49f962d3d434f944fd70a15c6ed12529f65b/app/views/change_requests/compare.slim), so maybe just double-check that we won't try to show the comparison on a change request whose proposed content has been removed.

mcrutch:
Confirmed - Approved change requests don't attempt to pull the compare.slim
Cancelled does not call compare either.

PR #509, opened 2021-03-25
Chagne the readonly text field into a span for display, makes it no l…

### LINKED ISSUE #340: Change Ownership modal's first input field should be unclickable and locked
Because users can't change it anyway, they shouldn't be able to click in it (they can't change it anyway). Take it un-tabable and unable to be focused CSS wise.

mcrutch:
…onger a tab-stop

Closes #340 

Thoughts on this?  It means it isn't greyed out, we could apply style to that to change it so it looks like the read-only field did, but it is no longer an actual field so isn't a tab-stop and no longer gives a mouse cursor indicator of being a form field either (changing to a text cursor over the entire field rather than just the text).


manfromjupyter (commented):
Your version is a lot less wordy for screenreaders and more ideal for ambulatory users once we add the CSS to make it look grayed out but the elements are reading the text together as one element (so in example below is reading it as "Current notebooschon smith"). We can move some elements around but I think we really just need to add a tabindex="-1" to the input field here. Could maybe also remove the type="text" to make it less noisy.

When it's set to readonly it doesn't even mention it's an input. I will try it with just the tabindex -1 and if it looks good for both use types of users, will apply it to the notebooks and resubmit it here. I think having in a readonly input is valuable because it lets you know the metadata and mirrors the structure of the value below you're replace it with. 
read only span - with edits

mcrutch:
You wouldn't use any value coming from that read-only field anyway because you have that exact data associated with the object already.

manfromjupyter (approved):

PR #519, opened 2021-03-30
Add the commontator views to the project for any customizations we ne…

### LINKED ISSUE #517: Comments have character limit of ~65k characters. Limit just to prevent breaks
Not many people even use the comment feature because provide feedback is more thorough, but if we're keeping it and not displaying it in another way users will use it more, need to just fix this issue for now to prevent potential 500 error.

mcrutch:
…ed to do

Set the maximum character length on comments to 60,000 on the form side
(Can't do anything serverside because it is all controlled by the gem)

Closes #517

### DIFF
--- a/app/views/commontator/comments/_actions.html.erb
+++ b/app/views/commontator/comments/_actions.html.erb
@@ -1,0 +1,29 @@
+<% # Clients of this partial must supply the following variables:
+   # comment
+   # user
+%>
+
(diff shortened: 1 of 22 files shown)


manfromjupyter (approved):
Works exactly as before except with the text limit added. Seems like a lot for this but it makes sense and does gets the job done. Thank you

PR #520, opened 2021-03-30
Fixed Code Cell Numbering for Pagination

### LINKED ISSUE #516: Code Cells on paginated notebook pages reflect links and coloring of wrong cells
Example, cell 21 (mighy translate to code cell 16) will show up as code_cell/1 if it's the top listing cell of page 2. The health and link is referring to that of code cell 1 (not 16).

Workaround: view notebook in full view.

manfromjupyter:
Closes #516 

Test by going to /notebooks/693-45-mixed-cell-notebook

Verify the link of the health icons consistently goes up based on what page and what cell you are looking at. Although we paginate after 20 cells, we paginate code cells based on only cells that are code ones, not "raw" or "markdown" cells.

### DIFF
--- a/app/views/application/_notebook.slim
+++ b/app/views/application/_notebook.slim
@@ -13,6 +13,31 @@ javascript:
   -cells = jn["cells"].paginate(page: @page, per_page: 20)
 -else
   -cells = jn["cells"]
+
+// Calculate what each page's code cell id and health should start at
+ruby:
+  page_cell_starts = [0]
+  if (cells.respond_to? :total_pages)
+    jn["cells"].each_with_index do |cell, index|
+      if index % 20 == 0 && index != 0
+        if cell["cell_type"] == "code" && !cell["source"].blank? && defined?(code_cells) && code_cells
+          code_cell_number += 1
+          page_cell_starts.push(code_cell_number)
+        else
+          page_cell_starts.push(code_cell_number)
+        end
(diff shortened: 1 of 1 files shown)


app/views/application/_notebook.slim
```diff
@@ -13,6 +13,31 @@ javascript:
...
+      elsif cell["cell_type"] == "code" && !cell["source"].blank? && defined?(code_cells) && code_cells
+        code_cell_number += 1
+      end
+    end
+  end
+  code_cell_number = 0
+
+// Designate starting code_cell_number based on page parameter
+-if (cells.respond_to? :total_pages) && params[:page] && params[:page].present?
+  -code_cell_number = page_cell_starts[params[:page].to_i - 1]
```
mcrutch, line 38: Why calculate the start index for every page and then lookup the appropriate cell number?  Why not just calculate the appropriate one, short circuit the loop and short circuit it so it stops counting so you aren't looping over the entire notebook once and then 20 rows of it.  This is just like, counting, but with extra steps ;)

mcrutch: which, I should ad, means this code shouldn't run 99% of the time, because 99% of the time people will only go to the first page and click Run, so why do all of this work?

manfromjupyter:
#521 is better. Closing this in place of that one

PR #524, opened 2021-03-31
Reload comments

### LINKED ISSUE #318: Posting comment does not cause page to reload thus having comment metric update for the notebook
Posting comment does not cause page to reload thus having comment metric update for the notebook (which makes people double guess if it really was added successfully).

Also remove the "#" that gets injected in the URL upon posting comment successfully (or perhaps on fails).

mcrutch:
Reload the notebook page after the comments dialog has been closed but ONLY if a new comment has been posted
Closes #318


manfromjupyter (approved):
Works appropriately

PR #527, opened 2021-04-01
Allow showing of private/deprecated in group display

### LINKED ISSUE #522: Support Toggling Private/Deprecated on Group Pages when all results are private and/or deprecated
Visit a group with just private and/or deprecated notebooks you don't own (ex. note to self, the "Drop Bearz" group). Verify the page loads as having no results but the table allowing you to toggle deprecated and private notebooks still appears so you can reveal them.

mcrutch:
When all notebooks are private/deprecated
Closes #522


manfromjupyter (approved):
Approved. Looks good. Thank you!

PR #528, opened 2021-04-01
Clean up emails to say when the "owner" didn't leave a comment

### LINKED ISSUE #523: Change Request Emails tweak when owner leaves no comment
Instead of saying `Owner comment: ""` when the comment field is left blank. Say either nothing or `Owner did not leave a comment.`

Also maybe "comments" should be "comment"? Idk

Current functionality (screenshot):
owner comments on change requests

mcrutch:
Also use .blank? rather than = nil to account for "       " type comments
Closes #523

### DIFF
--- a/app/views/change_request_mailer/accept.html.slim
+++ b/app/views/change_request_mailer/accept.html.slim
@@ -9,21 +9,23 @@ p
     - if !@owner.org.blank?
       |  (#{@owner.org})
 br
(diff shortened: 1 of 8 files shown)


manfromjupyter (approved):
Approved. Works well and confirmed with or without comments from the owner for all 3 types of emails both plaintext and html view work as expected.

PR #547, opened 2021-04-14
Fix an array index from "0" to "i"

### LINKED ISSUE #504: Environment Select from Run doesn't support opening in Lab
If you set an environment in LAB instead of classic and then set it as default, running in Jupyter will work but if you don't make any of them default and then manually select it when running, it won't work.

What looks like is happening is on the environments page the path is set to https://BLAH/jupyter/lab or https://BLAH/jupyter but when the little modal pops up asking which environment you want to run it in when one is not set as default, even if it's set to LAB, it will always only show the https://BLAH/jupyter format.

Stems from: #297 (merge #350)

mcrutch:
closes #504


manfromjupyter (approved):
Approved. Code looks good.

PR #548, opened 2021-04-15
Make the '+' icon for external resources ignored by screen readers

### LINKED ISSUE #537: External Resources icons need minor reformatting for accessibility
For add icon, they hear "Plus add new resource" instead of just "add new resource"  because the icon is being read for some reason. As for the trash/delete icon, it just appears but shouldn't when reading the text as an empty character "[]". Normally would just do an `aria-hidden="true"` on the icon, but we put the screenreader text directly under the icon. Minor but just wanted to flag it here before I forgot about it.

Currently
```
a.icon
  span.sr-only
```
Need to make it
```
a
    span.icon aria-hidden="true"
    span.sr-only
```
Might take some moving around of ids and classes to make it so the button still works.

mcrutch:
Closes #537

### DIFF
--- a/app/views/application/_external_resources.slim
+++ b/app/views/application/_external_resources.slim
@@ -2,7 +2,8 @@
   div.externalResources
     span.externalResourcesTitle #{GalleryConfig.external_resources_title}
     -if @user.can_edit?(@notebook) || @user.admin?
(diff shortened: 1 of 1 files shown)


manfromjupyter (approved):
Tested on screenreader too. Looks good.

PR #549, opened 2021-04-15
Date range search

### LINKED ISSUE #425: Add Search Filtering support for dates
Add abilities to filter results in searches using before: and after: to only find notebooks created or last edited before or after a given MM/DD/YYYY date (but support a change in configurations to switch to DD/MM/YYYY for Europeans, etc).

PS. Reason for MM/DD/YYYY instead of any other or dashes or anything is because for accessibility users that is how dates are portrayed in fields across the site already to date).

mcrutch:
Needs some accessibility testing/help from @manfromjupyter 
Closes #425


mcrutch:
Ok, no longer throws the 500 error for an invalid date, but I don't have a way to bubble up that error message right now.

mcrutch:
image
That massive gap can go away by taking out the no-header classes but I left it in for now in while you tinker with the accessibility in case you end up swapping around the visible text as well.

manfromjupyter:
UI fix for this for now.

When you first expand the Advanced Search, all you see is a dropdown field as shown:
Select Filter dropdown

Upon selecting the previous 8 options or the 2 new date time ones, you will see either of the following experinces:
created and updated shows appropriate fields
reveals rest after selecting

Other tweaks:
- when selecting "Add Filter", the value field will be emptied, but all inputs boxes will remain.
- when filling it all out but then change the filter type to another, will delete the value if it's a date-related field and wasn't before or visa versa, but otherwise will not affect the other fields (also replaces the middle input but that's implied).

manfromjupyter:
@mcrutch this is good to go on my end. Was distracted a bunch inbetween testing but pretty confident it's good to go. Please confirm this is the experience you want. Happy to test it again, just probably need you to confirm this is the experience you want. Fully accessible and as pretty as I could think to make it.

PR #558, opened 2021-05-19
Warning to site warning

manfromjupyter:
Closes #557

Issue #557, opened 2021-05-19
Change Warning database table to Site Warning


manfromjupyter:
One of the things hindering us from upgrading to a more recent version. Ruby 2.4 introduces a Warning module that is conflicting with our Warning activerecord.

PR #563, opened 2021-06-02
Ruby update to 2.4 (All that would be needed)

### LINKED ISSUE #556: Look into updating RDOC gem

manfromjupyter:
Then just need to:
- change version of ruby from 2.3 to 2.4
- install updated gems
- do a database migration (**bundle exec rake db:migrate**)

Closes #556 #557

INCLUDES Warning to Site Warning Pull Request (#558)


mcrutch:
What about updating the Dockerfile to ruby 2.4

PR #570, opened 2021-07-29
Utf8mb4

mcrutch:
Allow storage of 4 byte unicode characters in the database.  Using 🍍 as a test character, which does not work with the UTF8 storage, but does with utf8mb4


mcrutch:
Thinking about it some more, maybe I should just resize the columns to 191 vs resize the indexes..

mcrutch:
image
This was also storing the notebook itself in the database to test that unicode coming through. I used emoticons to make it noticeable in testing but things like mathematical symbols also can fall in the 4 byte Unicode range which is how I found this bug.

manfromjupyter (approved):
Code looks gucci

PR #574, opened 2021-07-30
Fix Star toggle

### LINKED ISSUE #572: star doesn't actually toggle
Click Star twice and both times it passes the same parameter.

mcrutch:
Closes #572 
Merge after #573


manfromjupyter (approved):
Looks good and works as expected

PR #584, opened 2021-08-13
Fixing User Summary

### LINKED ISSUE #575: User summary not refreshing for "All time" if setting any parameters prior
On the User Summary page, when you set the parameters on the date picker and then click "Submit", hitting "All time" after it refreshers does not reset the user data. Removing the dates found in the picker and then hitting "All time" also doesn't reset it.

mcrutch:
Closes #575
Fixes All Time button to work
Allows start and or end date to be empty both in support of All Time and also to allow the user to have an open ended date range
Converts date/times to iso8601 to ensure queries are successful

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -72,14 +72,14 @@ class UsersController < ApplicationController
   def summary
     min_date = params[:min_date]
     max_date = params[:max_date]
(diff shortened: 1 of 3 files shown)


manfromjupyter (approved):
Works perfect and code looks good

PR #585, opened 2021-08-13
Expose historical feedback to owners and editors of a notebook

### LINKED ISSUE #568: Provide Historical Feedback outside of email
Pain:
- User can't see provided feedback over time unless sifting through emails. 
- User that recently acquired a notebook from somebody else doesn't know of it's pending feedback so doesn't know the community's gripes with it or just ideas to improve.
- Comments which linger and can be seen over time do not send emails so owners are not notified with a new comment is made.

Proposed Solution(s):
- Send emails for comments too.
- When submitting provide feedback form, include additional field such as "Include as comment as well for historical record-keeping purposes". If they select that and agree that the request or message does not violate our terms of service, then it will include it as a comment as well.

mcrutch:
Closes #568

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -262,6 +262,8 @@ class ApplicationController < ActionController::Base
           title = "Revisions of \"#{@notebook.title}\""
         elsif url_check[3] == "reviews"
           title = "Reviews of \"#{@notebook.title}\""
+        elsif url_check[3] == "feedbacks"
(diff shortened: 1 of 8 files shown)


manfromjupyter (approved):
Code looks good, added Browser title, confirmed it works for accessibility users, confirmed json works, and and confirmed unauthenticated users (people lacking perms) can't see it (so the feedback is still private).

Issue #550, opened 2021-04-23
Add way to pull easy copy/paste list of all users that have run your notebook


manfromjupyter:
We currently have the metrics runs table but if we could have a way to easily copy/paste those emails without deleting the other columns rows manually and also having the ability to filter within certain dates, will allow them to accurately message their customers if certain information, say issues, or new features, etc. need to be communicated.

PR #586, opened 2021-08-13
JSON endpoints for components of the metrics page

### LINKED ISSUE #551: Add JSON endpoint for a notebook's runs, views, edits, stars, and edit history
Currently can get to /notebook/{id}/metrics.json thanks to the notebook controller's metric's method, but would be nice to break those returns into unique methods out of the notebook model so that we can return json object.

mcrutch:
Closes #551
Relates to #550 (allows you to pull the list programatically)


manfromjupyter (approved):
Works great, secure, and code looks gucci

Issue #439, opened 2021-01-22
Compliance reviews not handled properly across the site when enabled


mcrutch:
Reviews page gets wonky and it doesn't show up as a review available/completed on the right of a notebooks page under the dropdown menu.

(See pull request #435 for screen shot of the first problem)

PR #594, opened 2021-08-25
Responsive Tables

### LINKED ISSUE #535: Make all out-of-the-box tables responsive (Accessibility)
For 400% zoom support which is going to be requirement soon. Doesn't include presentation table for our notebook listings. Just reviews, revisions, change requests, metric tables, stages, /users, etc.

manfromjupyter:
Will eventually close #535 

Made environments table responsive and got everything ready to do the rest.

@mcrutch, can I have you checkout the Environment's Table at max zoom or with a mobile view and see if you are happy with this experience? If so I can roll it out to all of the others in no time at all (just might need to talk to you about what what columns we prefer be collapsed before others - as seen in the `gallery.js.erb` lines 1214-1216 (currently).

Will be sure to make everything else is supported at 400% zoom across the entire app, but tables are 95% of the problem at the moment.

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -1200,7 +1200,7 @@ $(document).ready(function(){
     },
   });
 
-  var changeRequestsAllTable = $('#changeRequestsAllTable').DataTable({
+  $('#changeRequestsAllTable').DataTable({
     'responsive': true,
     'paging': false,
     'language': { 'search': 'Filter results:' },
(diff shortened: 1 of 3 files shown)


mcrutch:
✨ Fancy ✨ 
Roll with it

manfromjupyter:
@mcrutch ready for you on this. Preformed all of the changes are mentioned here #535 and changed how we did the all reviews page to be efficiency O(n^2) instead of O(n^3) when we eventually add on compliance reviews. I can more confidently and quickly do #439 after this now.

PR #595, opened 2021-09-05
Implemented new user preferences setting to display full notebook cells

SweetMyCher:


mcrutch:
Functionally this works, but I would want to see a couple changes on this:
1. The user should not see this checkbox if GalleryConfig.pagination.paginate_large_notebooks is set to false.
2. Rather then "full_cells" disable_notebook_pagination feels like it fits more with our variable naming but @manfromjupyter might have a different idea on that will let him way in.
3. For the text on the checkbox in the user peferences page, I have a few thoughts running through my head, but right now the best is "Always display full notebook".  But again, @manfromjupyter please chime in.

manfromjupyter:
@RomanCherepov @SweetMyCher 
1. Agreed with @mcrutch (see above).
2. I have no preference, but the later does follow the verbiage we do tend to more learn towards (ex. "disable_notebook_cell_pagination"). Probably shouldn't just be "disable_notebook_pagination" because that's completely seperate to notebook cell pagination ;).
3. Agreed with @mcrutch (see above), "Enable full cells pagination" seems a little oxymoronic and should say ex: "Always display full notebook" or "Disable Notebook Cell Pagination"

PR #600, opened 2021-09-24
Hiding the link to change password when you aren't username login.  A…

### LINKED ISSUE #545: Users are provided with options to change password that have disabled the username_login_allowed config
On these 3 lines (249-251) on layout AND beta layout just need to be indented once.
https://github.com/nbgallery/nbgallery/blob/ddd297af9eaf03fb550b80089e73246a85fafcef/app/views/layouts/layout.slim#L248-L251 

In addition, also make sure on the /users/edit page that this whole section or page does not appear with that same config. User should still be bale to access /users/ID/edit, but not /users/edit to change their password, email, or cancel their account.

mcrutch:
…lready verified the change password screen is blank in this configuration
closes #545

### DIFF
--- a/app/views/layouts/beta_layout.slim
+++ b/app/views/layouts/beta_layout.slim
@@ -246,9 +246,9 @@ html lang="en"
                                   br
                                   strong ==@user.user_name
                               -if GalleryConfig.username_login_allowed == true
-                              li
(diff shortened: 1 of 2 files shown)


manfromjupyter (approved):
Good catch. Looks good.

PR #637, opened 2021-12-14
Tech Debt: Consolidated Space/Enter Key Support

### LINKED ISSUE #630: Tech Debt - Remove all of our individual Space or Enter key call outs
Will replace 161 lines of JS instead to just 8 lines.

Remove all of the individual space or keyboard checks to instead be, for example:

```
$('.keyboard-friendly, button').keypress(function(e){
    var keycode = (e.keyCode ? event.keyCode : event.which);
    if (keycode = '13' || keycode == '32'){
        e.preventDefault();
        $(this).click();
    }
});
```
And then add 'keyboard-friendly' class to all links or interactables that need it. Or can just make it 'a, button'  etc for the selector.

manfromjupyter:
Closes #630 

- Also fixed the calling of the login modal. Formerly was not reachable/callable but just keyboard. 
- Made some other selectors consistent with our style guide.

(thoroughly tested each element that was removed)


mcrutch:
Doesn't work on Notebooks pop-down since that is generated with an AJAX call (test by triggering the pop-down and trying space bar to activate one of the links)

mcrutch:
Ok, well, every other thing about the top bar activates with space (toggling the menus, activating the buttons, triggering search) but the Create Group button is the only button that doesn't work since that is in the AJAX dropdown

manfromjupyter:
@mcrutch, fixed merged conflicts here but getting a weird missing partial error for some reason on notebook pages. I will have to take a look unless you can figure out what the heck is happening to poor little ol me ^.^

mcrutch:
> @mcrutch, fixed merged conflicts here but getting a weird missing partial error for some reason on notebook pages. I will have to take a look unless you can figure out what the heck is happening to poor little ol me ^.^

app/views/application/_notebook_tile_small_title.slim is deleted in the most recent branch, that could be it? Digging deeper

Issue #202, opened 2020-06-30
Change HTML of alerts so that the Close is outside of the accessibility announcement.


manfromjupyter:
A little bit hard to explain but when an alert appears it interrupts the users due to changes that have been added for and if user is an accessibility user. Every time it announces it though it included the word "Close" at the end of it because we have the close icon in the same general container/block as the text that is being announced. Need to either move the close icon out or make the announcement `role="alert"` just around the text.

mcrutch:
Closing per comment in merge request #644

PR #644, opened 2021-12-14
Alert close button audit and type button outside forms and some htmlconsistency changes in our forms

### LINKED ISSUE #642: Tech Debt: Outside of forms, remove all type="buttons" from all buttons as it's pointless.

manfromjupyter:
Closes #642 and #202

- Did a button audit. Some <a> buttons injected role="button", removed type="button" for buttons outside of forms because I see no point, even for older browsers, and added aria-labels for a few that had text or a symbol for the stylesheet-less experience
- Moved the close button outside of the role="alert" container so that EVERY success and error message does NOT end with "Close".
- Moved the form element to start at the modal body, not encapsulate the entire form, including the header and modal close button. Really just making it consistent with the majority of our other forms.


manfromjupyter:
Marking WIP until I get time to fully test it again since we merged all of that stuff.

manfromjupyter:
@mcrutch, this is ready now. All retested

PR #305, opened 2020-09-18
Remove keywords

mcrutch:
Removes
- Keywords
- Wordclouds
- Imagemagick

Closes #297

PR #651, opened 2022-01-28
schema change to ensure correct build

JesperKrogh:
My first attempt to do a pull request - but the application does not build because of a missing keywords table in the setup.


mcrutch:
I see what happened.  It was MR #305 that deleted any references to it except this migration. Sorry about that.

db/migrate/20210726183045_change_to_utf8_mb4.rb
```diff
@@ -44,7 +44,7 @@ def self.up
     execute "ALTER TABLE groups CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
     execute "ALTER TABLE groups_users CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
     execute "ALTER TABLE identities CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
-    execute "ALTER TABLE keywords CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
+#    execute "ALTER TABLE keywords CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
```
mcrutch, line 47: Better to just nuke these two lines rather than level the references around.  Keywords are gone gone, and not coming back.

PR #660, opened 2022-05-20
All change request sorting

### LINKED ISSUE #619: Two fixes for Change_Requests/All
Currently  only affecting admins.

- Make change_request/All table a `.review-table` again
- Make first column of it sortable based on it's parsed table if the site has that configuration enabled

manfromjupyter:
Closes #619 

- Tried to make reviews and change request as close as possible with one another
- Added a data-sort to the change request column which I believe was what was undoing the sort we were adding lines before
- Added class to the tables that fixes styling

### DIFF
--- a/app/views/change_requests/all.slim
+++ b/app/views/change_requests/all.slim
@@ -24,10 +24,6 @@ div.content-container
           -sorted_array = @change_requests.order(:created_at).sort {|a,b| a.notebook.title.downcase <=> b.notebook.title.downcase }
         -sorted_array.each do |row|
           tr class="#{change_request_class(row.status)}"
-            -if GalleryConfig.notebook_title_parse.parsing_enabled
-              -notebook_title_sort = row.notebook.title.partition(GalleryConfig.notebook_title_parse.after_character)[2].downcase
-            -else
-              -notebook_title_sort = row.notebook.title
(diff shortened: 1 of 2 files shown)


manfromjupyter:
@mcrutch Reviews are disabled at the moment on my instance and the change_request/all was giving me a datatable alert for some reason. Just flagging those 2 gaps since fixing merge conflicts.

app/views/change_requests/all.slim
```diff
@@ -4,7 +4,7 @@ div.content-container
     div.center
       a.archived-option href="#{all_change_requests_path}?archived=true" tabindex="-1"
         button.btn.btn-primary Include Archived Requests
-  table.clean-table.responsive id=(@change_requests.empty? ? nil : "changeRequestsAllTable")
+  table.clean-table.review-table.responsive id=(@change_requests.empty? ? nil : "changeRequestsAllTable")
```
mcrutch, line 7: This is a change request table, not a review-table, but since both .review-table AND changeRequestsAllTable get the DataTable feature applied it was trying to doubly implement the change and thus the error

manfromjupyter:
@mcrutch, `review-table` is a styling, a type of table, that I think I was planning on needing to support that custom functionality to break have the multi-lined notebooks cells. Good to know that is what the issue is though. Will go back and find an intermediary solution if this alone didn't fix it.

PR #661, opened 2022-05-27
Feature to Simplify emails based on Condition

manfromjupyter:
**Included in this:**
- Added ability with a custom condition to simply the information shared over email.
- Fixed some punctuation and consistency issues between HTML emails and text version of emails.
- For accessibility, tweaks what text was the link so that users do not have to read the whole sentence to know what the link does.
- For feedback email, actually linked to our historical feedback page, which we didn't have before to link to.
- Similarly, for Accept and Decline change requests included link to all change requests for the user which before didn't exist because before we were purging change requests older than 7 days that were either accepted or declined.
- Subscription emails will be entirely simplified if any instance of needing simplification/redaction is found.
- Added config to force the making of all emails simplified.

## Screenshots of Simplified View
NOTE, included but not shown: notebook links only contain the ID, not the full name (ex. will link to the right place but only show `https://SITE_NAME/notebooks/345`).

### Create Change Request email
Create Change Request email

### Approve Change Request email (with and without comment)
Approve Change Request email - with comment
Approve Change Request email - no comment

### Cancel Change Request email (with and without comment)
Canceled Change Request email - with comment
Canceled Change Request email - no comment

### Decline Change Request email (with and without comment)
Declined Change Request email - with comment
Declined Change Request email - no comment

### Feedback email
Feedback email

### Share Notebook email
Notebook share email

### Subscription email
Subscription email - page 1

Subscription email - page 2

### DIFF
--- a/app/mailers/change_request_mailer.rb
+++ b/app/mailers/change_request_mailer.rb
@@ -4,7 +4,7 @@ class ChangeRequestMailer < ApplicationMailer
   def create(change_request, url)
     @change_request = change_request
     @url = url
-    @email_needs_to_be_simplified = email_simplify_check(change_request)
+    @email_needs_to_be_simplified = simplify_email(change_request)
     mail(
       bcc: @change_request.notebook.owner_email + [@change_request.requestor.email],
       subject: "NBGallery change request submitted"
@@ -15,7 +15,7 @@ class ChangeRequestMailer < ApplicationMailer
   def cancel(change_request, url)
     @change_request = change_request
     @url = url
-    @email_needs_to_be_simplified = email_simplify_check(change_request)
+    @email_needs_to_be_simplified = simplify_email(change_request)
     mail(
       bcc: @change_request.notebook.owner_email + [@change_request.requestor.email],
       subject: "NBGallery change request canceled"
@@ -27,7 +27,7 @@ class ChangeRequestMailer < ApplicationMailer
     @change_request = change_request
     @url = url
     @owner = owner
-    @email_needs_to_be_simplified = email_simplify_check(change_request)
+    @email_needs_to_be_simplified = simplify_email(change_request)
     mail(
       bcc: [@change_request.requestor.email, owner.email],
       subject: "NBGallery change request declined"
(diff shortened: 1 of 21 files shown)


manfromjupyter:
@mcrutch this is ready now! Screenshots shown

manfromjupyter:
@mcrutch Oh woops. Forgot one thing... to tweak the text-only email as well. Will do that first thing tomorrow but please let me know if there is anything you want me to change with current logic or any concerns you have.

app/mailers/change_request_mailer.rb
```diff
@@ -38,10 +39,20 @@ def accept(change_request, owner, url)
...
     @owner = owner
-
+    @email_needs_to_be_simplified = simplify_email(change_request)
     mail(
       bcc: [@change_request.requestor.email, owner.email],
       subject: "NBGallery change request accepted"
     )
   end
+
+  def simplify_email(change_request)
```
mcrutch, line 49: Why are you putting control logic in a view? Because that is how this code reads. Also, the name if this function implies that it is simplifying the email, not checking if it has to be simplified.

manfromjupyter: @mcrutch does this work? https://github.com/nbgallery/nbgallery/pull/661/commits/a14df830f22652cb4a59eec97dc7e3564d11193a

mcrutch: You are still using a (currently non-existent) view to make execution flow decisions, and there is no sample or documentation for how it would even work either.  Why can't this be a function or functions in the subscription controller than can be overwritten by an extension? Similar to how all the other custom permissions stuff occurs now.

manfromjupyter: Oh I see. My brain processed your earlier comment as only needing better naming

manfromjupyter: Made the change and tested it. Question, is this the way you want to do it?
```ruby
  def need_to_simplify_email?(change_request)
    if GalleryConfig.email.force_simplified_emails
      return true
    end
    return false
  end
  ```
  And then have a custom method that overrides this? I only ask because to people looking in, we have parameters that are never used.

File: app/views/notebook_mailer/share.text.slim [hunk 2]
```diff
@@ -4,6 +4,7 @@
 - if !@owner.org.blank?
   '  (#{@owner.org})
 ' has shared with you the notebook #{@notebook.title}: #{@url + notebook_path(@notebook)}
```
mcrutch, line 6: You need to sanitize the title here like you do with the HTML version

mcrutch (commented):
So far I have found that this doesn't cover comments being posted, which is a problem and will be difficult to find, but I also wanted the ability to extend this better,
So......
take a look at my branch of this: https://github.com/nbgallery/nbgallery/tree/email-summaries-mcrutch  I didn't push directly to your branch because I want it to be easier to diff and adapted what changes I made to what you are working with.

Still doesn't handle the real-time comments  and I haven't gotten through testing everything (testing is a PAIN especially subscriptions) but there are some big things that need to change and should change.

PR #665, opened 2022-05-29
Quick Copy Buttons on Metric pages for all User Emails

### LINKED ISSUE #550: Add way to pull easy copy/paste list of all users that have run your notebook

manfromjupyter:
Added buttons on bottom of metric modals to quick copy all of the user emails.

Closes #550. The only thing is this only works with HTTPS websites. The old method to do it for HTTP sites requires old browsers which we block. Maybe there is another way, but this seems to be the only way.

### DIFF
--- a/app/assets/javascripts/notebook_actions.js.erb
+++ b/app/assets/javascripts/notebook_actions.js.erb
@@ -745,4 +745,21 @@ $(document).ready(function(){
       element.style.cssText = 'height:' + value + 'px';
     },0);
   }
+
+  /* ===================================== */
+  /* ========= Copy to Clipboard ========= */
+  /* ===================================== */
+  $('.copy-to-clipboard').click(function() {
+    if (navigator.clipboard && window.isSecureContext){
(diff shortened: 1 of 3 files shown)


mcrutch:
We do not expose email addresses of the users to anyone other than the admins anywhere on the site. (unless the configuration has the username and email the same).  So I am not willing to merge this unless it changes to user_name versus email (which in turn potentially limits its functionality) but we are not going to expose all the email addresses of everyone who has run the notebook.

manfromjupyter:
@mcrutch, oh good catch. I didn't think about that. I can change it to that. If the people's usernames are the prefix of their email within an organization, pasting that in an email to field and having the email app check names can populate emails if it exists. I see it as a way to fulfill our security requirement and give this functionality. If you disagree though, we can shelve this.

PR #672, opened 2022-07-22
Make NBGallery tolerant to SOLR failures

### LINKED ISSUE #616: make nbgallery fail gracefully if SOLR is unavailable
If for some reason SOLR is unavailable OR notebook indexing fails...fail gracefully.
Save notebook updates without indexing it in the search (can reindex later)
If user attempts to search present a message that says "Search is currently unavailable" or something similar rather than the scary 500 error page.

SOLR is the hardest component to make fault tolerant in a production environment, so instead make the app fault tolerant of no solr.

mcrutch:
For sanity sake I want a few more sets of eyes.  I stopped the auto-indexing so the save to the database can succeed, have an aftersave that will attempt to index the records (for both groups and users).  I am not happy with how the error message appears on the search page so I think @manfromjupyter should double check that I set the flash message correctly (it stays one screen past when the user hit the error message on a bad search)
Closes #616

### DIFF
--- a/app/controllers/notebooks_controller.rb
+++ b/app/controllers/notebooks_controller.rb
@@ -747,19 +747,28 @@ class NotebooksController < ApplicationController
   # GET /notebooks
   def index
     @notebooks = query_notebooks
-    if params[:q].blank?
-      if !params.has_key?(:q)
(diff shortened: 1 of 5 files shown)


jeffyjefflabs (approved):
Looks good to me

PR #674, opened 2022-07-27
Make search for users and groups during change ownership case-insensitive

### LINKED ISSUE #673: Make Change owner search drop-down case insensitive

mcrutch:
Makes Group and User auto-complete during change ownership case insensitive
Closes #673


manfromjupyter (approved):
Ugggggh, fine

PR #678, opened 2022-07-27
Pass full metadata on download

### LINKED ISSUE #567: Include more metadata when downloading notebooks
Currently, when we download a notebook, we insert some gallery metadata into the metdata section of the notebook's json ([code](https://github.com/nbgallery/nbgallery/blob/main/app/controllers/notebooks_controller.rb#L266-L274)). This is basically just the minimal set of ids and whatnot neccessary for a Jupyter-nbgallery integration to work. Proposal for ticket is to instead insert the rest of gallery metadata that is available from the metadata endpoint ([code](https://github.com/nbgallery/nbgallery/blob/main/app/controllers/notebooks_controller.rb#L222-L252)).

mcrutch:
Pass all attributes as part of the gallery metadata on download.  This only adds fields it does not affect any that are already there
Closes #567

### DIFF
--- a/app/controllers/notebooks_controller.rb
+++ b/app/controllers/notebooks_controller.rb
@@ -318,7 +318,32 @@ class NotebooksController < ApplicationController
     jn = @notebook.notebook
     jn['metadata'] ||= {}
     gallery = jn['metadata']['gallery'] ||= {}
-    gallery['uuid'] = @notebook.uuid
+    Notebook.attribute_names.each do |attr|
+      case attr
+      when 'owner_id'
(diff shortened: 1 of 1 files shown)


manfromjupyter (approved):
Oh this is genius! Tested it and it works perfectly.

Did we want to maybe add to the default description on upload of these something like "Fork of John Smith's notebook 'Testing with Potatoes'"? and/or include link URL link to it in the description?

mcrutch:
We strip gallery metadata on upload so passing it up wouldn't happen

PR #680, opened 2022-07-27
Allow for creation of a description template

### LINKED ISSUE #531: Look into adding dependencies section within Notebook metadata
Need to bake it out more what exact kind of thing would be included in here. If less than 10% of notebooks would benefit form this, I don't see a benefit to it but yolo. Without much context, this is what has been imagined. A metadata section titled; "Dependencies:" followed up by a tag-like list of elements such as.... idk.... "requires EXCEL fil for upload", "required training X", required access to Marketing Database", etc.

OR could add it as a sections to the bottom of the notebook jumbotron that expands/collapses. Ex. move comments down to a tab dropdown, move Recommended notebooks over, add a Dependencies section, etc.

Original question: "It would be nice to add a bit more structure of optional fields to the description of the notebook. Having a specific place for people to list things like accesses to other departments or what databases to be able to run the notebook, if you need to upload a file, and if so what format in order for this to work. This would make it easier to see if a notebook is likely going to work for me."

mcrutch:
Closes #531

@manfromjupyter I think we should increase the default size of the description input (height) or have it auto scale, because the second you shove a description template in there it is just lost in the noise with the scroll bar but the box doesn't enlarge itself.  Can you take a look into that.

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -985,6 +985,15 @@ $(document).ready(function(){
           $('#stageUpload').modal('show');
           $('#uploadFileModal').modal("hide");
           $('#uploadFileSubmit').attr('disabled', false);
+          var descriptionTextarea = $('#stageDescription');
+          var checkModalRendered = setInterval(function() {
+            if (descriptionTextarea[0].scrollHeight > 1) {
+              document.addEventListener('keydown', autoSize);
+              let value = descriptionTextarea[0].scrollHeight + 2;
+              descriptionTextarea.css("height", value + "px");
+              clearInterval(checkModalRendered);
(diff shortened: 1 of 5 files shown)


manfromjupyter:
Agreed. Will look into expanding and fire back with some options.

manfromjupyter:
@mcrutch 
- Modal now renders with height dynamically set to how big the description template is.
- Description and title fields now have backup placeholder text on the Notebook page if person deletes the title or description (and walks away, it saves for some ungodly reason, or whatever else). Accessibility thing
- Set min height on notebook description field just because you could previously make it like 3 pixels tall making it slightly hard to reopen again.

manfromjupyter (approved):

PR #682, opened 2022-07-27
Allow admins to turn off HTML descriptions

### LINKED ISSUE #650: Long Descriptions with HTML truncates weird
TODO: See if it can be reproduced

What was seen:
"If a user puts HTML in the description of their notebook and it's a long description when the notebook is shown in a search result, the description gets cut short which can leave tags open. This then plays havoc with the rendering of the page and can result in parts of different notebooks being stiched together to appear as one, or the tags and stats for the notebook to just not be rendered."

### LINKED ISSUE #681: Add option to turn off HTML descriptions, pass all display of description/title through a sanitization function on output

mcrutch:
Alternative way to Close #650
Closes #681

### DIFF
--- a/app/views/application/_notebook_tile_small.slim
+++ b/app/views/application/_notebook_tile_small.slim
@@ -16,7 +16,7 @@ div.info.square
         '  Notebook description:
       a.nounderline.notebook-description.tooltips href="#{notebook_path(notebook)}" title="#{notebook.description}"
         i.glyphicon.glyphicon-info-sign aria-hidden="true"
-        span.sr-only ==notebook.description
+        span.sr-only ==strip_tags(notebook.description)
       -if defined? last_viewed
         -if notebook.updated_at > last_viewed
           i.glyphicon.glyphicon-fire.tooltips tabindex="0" title="Notebook has been updated since being viewed last" aria-hidden="true"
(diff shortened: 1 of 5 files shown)


app/views/application/_notebook_tile_small.slim
```diff
@@ -16,7 +16,7 @@ div.info.square
         '  Notebook description:
       a.nounderline.notebook-description.tooltips href="#{notebook_path(notebook)}" title="#{notebook.description}"
         i.glyphicon.glyphicon-info-sign aria-hidden="true"
-        span.sr-only ==notebook.description
+        span.sr-only ==strip_tags(notebook.description)
```
mcrutch, line 19: This is for screen readers only so wanted to strip tags but not "render" it to possibly break things

config/settings.yml
```diff
@@ -151,6 +151,7 @@ video:
 
 # Notebook entries
 notebooks:
+  allow_html_description: false
```
mcrutch, line 154: defaulted to false because this is a MUCH safer mode of operation

manfromjupyter:
@mcrutch, how do admins disable the HTML descriptions or is this only for system admins that can add the environment variable?

mcrutch:
Only SA's not UI administrators

manfromjupyter (approved):
Code looks good

PR #688, opened 2022-08-15
User Forms Refactor

### LINKED ISSUE #687: Resolve Browser Warnings Dev Tools warnings regarding passwords/emails
For the half dozen forms we have with passwords and emails, need to make sure these IDs are unique, that autocomplete is correct, types are correct, and perhaps take the time to revisit any accessibility issues.

manfromjupyter:
Closes #687, #691
- **Feature:** Fixed autocomplete, ids, and types of various user form fields necessary to dismiss the various Google Chrome dev tools warnings.
- **Feature:** Redesigned the user forms to closer mimic that of our forms found on modals (with exception of main Login, Sign up, and Register forms in modals - these changes only touch their inlined-form equivalent).
- **Feature:** Added new forward link to User Edit page to link user back to their Notebooks page (with all of the other helpful user links if user was not coming from that page). Also added forward link to Change Password page so user has easy link to change all aspects of their account should they desire it.
- **Bug fix:** Fixed a handful of accessibility issues (types, labels, headings, etc.).
- **Bug fix:** Fixed users/unlock/new page which had a broken UI.
- **Bug fix:** Fixed issue with all of the devise generated emails where a space character was missing between "Hi" and the user's username/email.
- **Tech Debt:** Removed as much of the "Rails" logic for generating these forms in favor of the standard HTML way we do to generate all of the others.
- **Tech Debt:** Relabeled some classes such as return (back) link and the input class to add a min width to make it more clear to other developers in the future.

## Screenshots
Sign Up when inlined
Login form when inlined
Resend Confirmation Instructions form
Edit User form
Change Password page
NOTE previous screenshot shows Changing Password while user is signed in. Following screenshot shows Changing Password from email.
Change your password page (from email)
Resend Unlock Instructions form

Forgot your password form

### DIFF
--- a/app/assets/javascripts/federated_search.js
+++ b/app/assets/javascripts/federated_search.js
@@ -49,7 +49,7 @@ function ajaxSearchError(tabElem, response) {
   tabElem.find(".logo-loading").addClass("hidden");
   tabElem.find(".tagline").addClass("hidden");
   tabElem.find(".tab-pane-content").css("opacity", "1");
-  tabElem.find(".result-container").html('<div class="alert alert-danger">There was an error obtaining search results.</div>');
+  tabElem.find(".result-container").html('<div role="alert" class="alert alert-danger">There was an error obtaining search results.</div>');
 }
 
 function updateHistory(url, tabId) {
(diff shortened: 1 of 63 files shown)


manfromjupyter:
Ready to merge again

PR #694, opened 2022-09-22
Feature to enable Simplified Emails

manfromjupyter:


manfromjupyter:
@mcrutch whenever I try to run the subscription mailer for whatever reason (before and after doing my latest change) I get the following error:

cant test subscription emails  Argument error

"wrong number of arguments (given 1, expected 2)".

It's super weird because there are 2 there. If I separate with a space, it still says 1, BUT if I add another parameter like "hello", it says given 3 arguments, expected 2 lmao. Anyway, do you see something I don't here?

mcrutch:
The problem was with the simplify_email? function.  I added a default for message, try things now.

manfromjupyter:
@mcrutch, this is ready now. Verified it works for all types of emails. Screenshots of these are found in the other PR that is now closed. Also with this ticket I added the functionality to support Compliance reviews in subscription emails and also fixed an issue where some NBG environments have an ending slash with the `@url` element and because our paths start with a slash for all of these notebooks, users, groups, etc, the url was looking site `https://site_name//notebook/123`. Added some logic to clean that up IF the owner of a given NBG adds an ending slash in the config or whatever.

PR #710, opened 2022-12-02
Upgrade/rails to 5.2.8

ScratZa:
Draft PR 

Devise seems like its not working but should  figure we should use 4.0 - still think it will be broken though 
Seems like the indexing on a fresh database breaks when running a migration - should probably look into that
No idea what else is broken but at least managed to get it up and running (hacky sim links for nmatrix on bundle install was the only way i could get it to work)


Gemfile
```diff
@@ -73,7 +72,7 @@ gem 'sunspot_solr'
 
 # Nightly computation - similarities, suggestions, etc.
 gem 'activerecord-import'
-gem 'nmatrix'
+#gem 'nmatrix'
```
ScratZa, line 75: re enable this

app/controllers/commontator/comments_controller.rb~
```diff
@@ -0,0 +1,160 @@
+module Commontator
```
ScratZa, line 1: Remove this file

app/controllers/commontator/threads_controller.rb~
```diff
@@ -0,0 +1,75 @@
+module Commontator
```
ScratZa, line 1: remove this file

mcrutch (commented):
Still working through this, but made some changes to make it work better.  Wanted to leave this now in case you had time to pick it up before I got back to it.

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,6 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   layout 'layout.slim'
-  protect_from_forgery with: :exception
```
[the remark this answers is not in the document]
ScratZa: i believe it is on by default from 5.2 provided the 5.2 defaults are loaded 
https://github.com/rails/rails/commit/ec4a836919c021c0a5cf9ebeebb4db5e02104a55

ScratZa:
> Still working through this, but made some changes to make it work better. Wanted to leave this now in case you had time to pick it up before I got back to it.

awsome yea thanks a bunch for the help on this - first time ive really done any work on ruby so could be missing quite a bit

app/controllers/application_controller.rb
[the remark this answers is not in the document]
ScratZa: https://railsdiff.org/4.2.11.3/5.2.8.1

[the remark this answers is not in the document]
mcrutch: Yep, confirmed on by default in 5.2+
https://edgeguides.rubyonrails.org/configuring.html#config-action-controller-default-protect-from-forgery

app/mailers/commontator/subscriptions_mailer.rb
```diff
@@ -0,0 +1,49 @@
+module Commontator
+  class SubscriptionsMailer < ActionMailer::Base
```
mcrutch, line 2: @manfromjupyter We will need to evaluate this with the simplify email stuff.

db/migrate/20210726183045_change_to_utf8_mb4.rb
```diff
@@ -74,68 +44,8 @@ def self.up
     execute "ALTER TABLE users CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
     execute "ALTER TABLE users_also_views CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
     execute "ALTER TABLE votes CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
-
```
mcrutch, line 77: All of these execs are because old copies of the database were just 2 byte utf8 and so the indexes had to be removed before the columns were converted to a better character set, and then added back (but with a shortened key length due to mysql limitations)  Not sure why you deleted all of these. there had been a small problem with this migration in the past but that was fixed.

mcrutch:
Been having to dig through this a lot but making a lot of progress.  Rails 5 really threw somethings for a loop with this PR.  @manfromjupyter and I have been hammering on it with our test instance, and I feel like we are getting close on this.  Subscription emails work again, groups can be created again, but due to problems on my end I haven't been able to test oauth properly yet and I want to get that tested as there were some changes in the oauth plugin.

ScratZa:
Awesome - Yea sorry chaps wish I could've figured out a lot more of this for you - as I mentioned not familiar with ruby and rails so this was an experience to say the least :) really appreciate it  - learning a lot as well. Haven't had much time on my hands to go through everything just yet but shouldn't be the case from next week onwards.

mcrutch:
Items left to review that I can think of
- [ ] Reviews
- [ ] Group Modification
- [ ] OAuth
- [ ] Change Requests
- [ ] new revisions
- [ ] revision history
- [ ] saving Jupyter Lab preferences
- [ ] NBGallery Preferences
- [ ] Upload new notebook
- [ ] Nightly computation scripts
- [ ] Change ownership to user
- [ ] Change ownership to group
- [ ] Gear menu on notebook page items
- [ ] Provide Feedback
- [ ] Comments and related notifications (simplify email comes into play here)
- [ ] Simplify email as a whole
- [ ] Deprecation

manfromjupyter:
Fixed an issue breaking Reviews page that was left 5 months ago (just needs to be confirmed tomorrow probably). Also, know this probably should have been placed in another ticket/PR but fixed a styling issue that's been eluding me for years where health icon sometimes appears in center of notebooks if the notebook was created through unconventional means. Fully tested. It's harmless, just some CSS and added a container around existing code cell health icons.

app/views/change_request_mailer/accept.text.slim
```diff
@@ -1,9 +1,9 @@
 ' Change Request Approved
 = "\r\n" * 2
 -if @email_needs_to_be_simplified
-  ' Change Request for your Notebook: #{@url + notebook_path(@change_request.notebook).split('-')[0]}
+  ' Change Request for your Notebook: #{notebook_url(@change_request.notebook).split('-')[0]}
```
mcrutch, line 4: Work this one as well

mcrutch:
oauth stuff is all broken at the moment. Looking at updated gems

mcrutch:
Found a problem.  Mathjax-rails is not working with Rails 5.2 because of hte default CSRF protection thinking it is a call from an external site.  WHY?!?!?!  who knows.  Digging into this.

mcrutch:
Can't find an easy way to make rails-mathjax to work with Rails 5.2 and it hasn't been updated in YEARS.  might need to pull it into our asset pipeline?  I hate asset pipelines

mcrutch:
Other than Azure (which at this point isn't exposed to the user for logging in anyway) I think we are actually good to squash/merge this and then clean up any small issues that pop up after the fact.  I even merged this into our closed instance with some extra extensions and everything seemed to be rolling fine.

Sorry I had to merge bomb the mathjax stuff in, only way I could reliably get it to work since the gem is so janky.

PR #718, opened 2022-12-20
Custom url

mcrutch:
Allow NBGallery to be run at a sub-path on a server rather than /


app/views/notebooks/index.rss.builder
```diff
@@ -26,7 +26,7 @@ xml.rss(version: '2.0') do |rss| # rubocop: disable Metrics/BlockLength
       channel.item do |item|
         item.title nb.title
         item.description nb.description
-        item.link "#{request.base_url}#{notebook_path(nb, ref: :rss)}"
```
mcrutch, line 29: This is the only one of these _path switchouts that was truly necessary. Going to roll back the rest most likely

app/assets/javascripts/gallery.js.erb
```diff
@@ -203,7 +203,7 @@ $(document).ready(function(){
     $("#newHomePageNotebooks").css("opacity", ".5");
     $("wait").css("display", "block");
     loadingGif();
-    tabUrl = '/home_notebooks';
+    tabUrl = application_base_url +'/home_notebooks';
```
manfromjupyter, line 206: Maybe want a space after the `+`?

manfromjupyter (approved):
All looks good except the missed space and if you wanted to change back all of the `noteback_urls` to `notebook_paths` except the one you mentioned.

mcrutch:
There were one or two features that still didn't work that I have to look into.  I have them in testing notes from @blueninja416

PR #725, opened 2022-12-22
Improve json endpoints

### LINKED ISSUE #649: Support JSON REST Api calls for each page
Currently we offer an HTML, text-only view of most every page, but this would be to actually render in HTML.

Example https://{SITE}/notebooks/{ID}/metrics.json is in JSON

Example [h](https://{SITE}/notebooks/{ID}/revisions.json is in HTML. This is what we want to be made JSON.

mcrutch:
Add appropriate handlers for critical .json endpoints for better external visibility Not touching reviews, user_preferences or admin pages intentionally
Closes #649


manfromjupyter (approved):
Flawless

PR #731, opened 2022-12-22
Improve error message when public/private toggle fails

### LINKED ISSUE #611: Making a notebook public/private that has illegal characters in not returning proper error message
Error message received was "You had an error" or something similar like "Error: Error". Need to make sure we return a JSON response or whatever the issue maybe be. Checking the network protocols shows the message was in the response but didn't get injected somehow.

mcrutch:
Closes #611


manfromjupyter (approved):

manfromjupyter (changes requested):
The alert message you produce says problem with the description.

mcrutch:
Oops, copy/pasta, will fix that before mege

PR #732, opened 2022-12-22
Use MakeAlert to display a message when you star/unstar a notebook

### LINKED ISSUE #220: Add success banner/message when user stars a notebook
Left to do from #183. For accessibility reasons, the private/public work just fine as well as subscribing because the success message is a role="alert" which lets screenreading users know when their action worked, we need the same for starring.

mcrutch:
Closes #220


manfromjupyter (approved):
Looks good

PR #733, opened 2022-12-22
Made it so inline errors appear when not deprecated

manfromjupyter:
Haven't been table to test to see if alerts work when the page is deprecated but when not deprecated it works now.. Can confirm.


mcrutch:
This is closer to what it needs to be but there is still some CSS funkiness going on.  Top is without deprecation, bottom is with deprecation.
image

manfromjupyter:
@mcrutch, fixed now finally, Geez.

PR #735, opened 2022-12-23
Git rid of NMatrix because it brings in backports which generates a warning in rails 5.2+

### LINKED ISSUE #729: Investigate replacing nmatrix due to use of deprecated ruby features

mcrutch:
Recommendation scores are super close, density numbers are off on arrays but that doesn't really matter that much.
Need to test this on a large data set to see what happens memory wise.
Closes #729


mcrutch:
@jeffyjefflabs - I replicated the functionality with similar functions in Matrix to NMatrix but the comments talk about being able to do it more easily with a fully populated matrix, which this provides, so if you could provide any insight into that, it would be appreciated.

jeffyjefflabs:
If the final scores are close then it's probably fine.  The density numbers aren't really important; I was just logging them to see how much memory the sparse matrix was saving, since we were originally very constrained.  If you're not using a sparse matrix, then I wouldn't even bother computing them.  (I vaguely recall that the sparse matrix "size" might include some of the bookkeeping info in addition to the non-zero elements, so that might explain why your computed value is different.)

I think the comments you're referring to are for the Jaccard similarity computation.  There's a way to do that with matrix multiplication like we're doing for the other scores.  The problem (given our original memory constraints) is that it requires an all-ones matrix (the "fully dense matrix" referred to in the comments) - so that immediately cancels any memory benefit from the sparse matrix of the intersections.  Here is the [best reference](https://royhung.com/jaccard-index) I could find quickly - basically our current code computes the intersection with matrices as described there, but then we do the union and the final score row-by-row so it's not all in memory at once.

jeffyjefflabs (approved):
Looks good to me; see comment in discussion thread for historical background.

mcrutch:
This doesn't fall over until data sets get really large.  Probably worth just merging this and providing guidance for how to work around it for large data sets (I have a python version written for the two matrix jobs)

PR #739, opened 2022-12-23
Emails to Alert Owner of Notebook on Share/Ownership changes

### LINKED ISSUE #455: When a notebook is shared by someone other than owner, notify owner
**Low priority** (doesn't do it now anyway) but would be a nice to have down the road. 

The email could be as simple as 
> subject: "A NBGallery Notebook you own has been shared"
>
> "***username*** has shared the notebook ***notebook*** with x people."

Maybe add instructions to see the share list by clicking share notebook?

Only done for notebooks owned by an individual user.

manfromjupyter:
- Made messaging for what sharing is consistent
- Added more clarity to variable names for shares
- Created separate email to alert owners when their shared users share their notebook with others.
- Add alert email for ownership change emails. Currently only admins can change the ownership out from under someone but still seems prudent.
- Did some consistency changes on all emails.
- Added underlines to all links to make them stand out more (accessibility improvement too).

Closes #455


manfromjupyter:
The two new emails:

share with others email
ownership change

mcrutch:
Rather than "Added or removed a shared editor" how about
"changed who the notebook is shared with"

manfromjupyter:
@mcrutch, this is ready again (plus merged latest). Only thing it doesn't have is the second double-alert fix.

PR #740, opened 2022-12-27
Ensures that as long as a tag exists, the user will remain subscribed…

### LINKED ISSUE #353: Subscription Page 404ing when notebook, tag, or group record deleted

mcrutch:
… to it

Moves the subscription from the deleted instance of a tag to another Closes #353 
If all instances of a tag are deleted, the subscription goes away, but this is an acceptable trade-off to the alternatives

#360 should actually still be handled better, but this is a stop-gap

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -2,7 +2,19 @@
 class Tag < ApplicationRecord
   belongs_to :user
   belongs_to :notebook
-  has_many :subscriptions, as: :sub, dependent: :destroy
+  has_many :subscriptions, as: :sub
+  before_destroy  { |tag|
+    subscriptions = Subscription.where(sub_type: "tag").where(sub_id: tag.id)
+    tag = Tag.where(tag: tag.tag).where("id != ?",tag.id).first
+    if(tag)
+      subscriptions.each do |subscription|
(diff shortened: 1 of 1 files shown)


manfromjupyter (approved):
@mcrutch, Only potential issue I see is it takes probably 10 seconds to remove the tag from a notebook and potentially more on a non-dev system. This lag might be confusing to people because if any user is subscribed to the tag it will just lag a bit for them when they try deleting. Still, the fact it works and no longer breaks on delete is huge, but if there was a way to make the subscription and tag lookups faster, it would be quite valuable.

mcrutch:
Our dev site has been extra slow for one, mostly because of super excessive logging, and it only has to make changes if that specific tag instance has subscriptions and only extra work if that tag also exists on other notebooks.

PR #742, opened 2022-12-28
Make the usage of "tag" less ambiguous throughout the code

### LINKED ISSUE #463: Fix ambiguity of the variable name "tag" through the code
Sometimes tag is used to store a tag object
Sometimes tag is used to store the tag text
Both uses are often found in the same file (see daily_subscription_email_text.slim as an example)

Tracing bugs involving tags is very difficult because of this and it has caused problems when doing a copy/paste between views in the past.

mcrutch:
Prepare codebase for normalization of the tag field Closes #463
Related to #360

### DIFF
--- a/app/assets/javascripts/notebook_actions.js.erb
+++ b/app/assets/javascripts/notebook_actions.js.erb
@@ -110,7 +110,22 @@ $(document).ready(function(){
         location.reload();
       },
       error: function(response){
-        makeAlert('error', '' , 'System encountered an error when trying to save notebook tags. ' + response.responseText);
+        if (response.responseJSON["errors"] && Object.keys(response.responseJSON["errors"]).length > 0){
+          errorMessage = "";
+          console.log(response.responseJSON["errors"]);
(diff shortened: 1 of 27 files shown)


manfromjupyter:
Still testing this but did we also want to rename "tag" in the SQL table? 

tag in table

mcrutch:
Longer term, that is going away so didn't touch it.

manfromjupyter:
Okie doke. Only other problem is that `/tags` page doesn't render due to an error. I will fix this though

tags page error

mcrutch:
since that is an S and assuming that is the only issue, I am going to guess I was trying to hit Ctrl+S with the character highlighted at some point and didn't retest that page, because I had that page working

manfromjupyter (approved):
Wow that took me slightly too long to figure out even after you saying "S" was the issue hahaha. Confirmed that was all that was needed and everything else looks good. Search worked for tag filtering also but maybe I also want to flag that after doing it twice no searches at all work. Maybe just a SOLR issue?

PR #744, opened 2022-12-29
Add downloads to the notebook metric page

### LINKED ISSUE #193: Show downloads on notebook metrics page
We are tracking this in the database (last I checked), we just need to reveal it.

mcrutch:
Open to ideas for changing color but downloads are similar to runs so green for now 
Closes #193

image


manfromjupyter (approved):
@mcrutch, thank you fixing my copy/paste error with the screenreader only text. As per the downloads, nice work getting that metric built out. Per making it green I was conflicted at first but I've come around over the last little bit looking it and thinking of alternatives; it mostly makes sense to marry the two as they are the both show the value of your notebook and also supports the usecase of those that download to use. 

Only thing I don't like is the bootstrap styling but I recognize that's been there a while.

PR #749, opened 2022-12-30
Limit length of group and notebook names

### LINKED ISSUE #748: NBGallery doesn't properly handle titles that are too long
Rather than gracefully failing with a warning it throws a 500 error if you enter a title with too many characters

mcrutch:
group names limited to 100 to prevent overflow problem
Closes #748

### DIFF
--- a/app/assets/javascripts/groups.js
+++ b/app/assets/javascripts/groups.js
@@ -67,7 +67,7 @@ $(document).ready(function() {
 
     element.addEventListener('ajax:error', function (event){
       [data, status, xhr] = event.detail;
-      makeAlert('error', '#groupForm .alert-container' , 'Group creation failed: ' + cleanJSON(data));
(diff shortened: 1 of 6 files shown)


manfromjupyter (approved):
Confirmed it works and approved but would you like to also add an ex: `maxlength="250" to these and the little `span.remaining-characters-warning` we use on change logs and deprecation messages to show how many characters they have remaining before they submit?

mcrutch:
I am fine with that getting added.  Do you want to throw it on there?

manfromjupyter (approved):
I'm happy with this now. See screenshots; both have the limits set and have the character counts enforcing it. Only thing slightly off is notebook title is not auto-expanding as I type but I'm unable to get it to work in a timely manner so going to just leave it as is.

notebook title character checl
groups title character check

PR #756, opened 2023-01-06
Step 1 in removing the will_paginate-bootstrap

mcrutch:
Now the pagination links must be styled again


manfromjupyter (approved):
Looks good for me, confirmed params work for it, confirmed it works on hover/focus, and being on first or last causes the next/previous button to be black and unclickable. Only thing different is to blind users won't read "listing of 8 links" if it has 6 pages (and next/previous button), it will just read navigation region and then read the links in order. In my opinion this probably preferred anyway despite it not saying it's a list of links on screenreaders.

PR #761, opened 2023-01-06
Make sure the last tab that was clicked on is the one that sticks

### LINKED ISSUE #760: Get rid of race condition on loading homepage tabs
Click a slow tab (like "all" or "suggested" then click a fast tab like "stars" and the stars tab will appear than get overwritten by the slow tab.

mcrutch:
Also centralize most of the tab loading code to one function
Closes #760 
Activating tabs by click, enter key and space still works too

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -215,11 +215,9 @@ $(document).ready(function(){
     currentTab=tab;
     $("#txt").css("opacity", ".5");
     $("#newHomePageNotebooks").css("opacity", ".5");
-    if ($('#initialLoadImage').length < 1){
-      $("wait").css("display", "block");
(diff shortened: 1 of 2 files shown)


manfromjupyter:
Only issue I see is on initial load (or after you "Empty Cache and Hard Reload"), the nbgallery loading icon is also transparent but not after clicking on any tab. I think the 0.5 opacity needs to go back to what it was. The default listings have opacity already so users should hopefully know already when its in-between tabs/loading.

mcrutch:
Having a loading spinner in the text + one displayed from the javascript was really screwy.  Got rid of the one in the .slim and now always show the other one, and that solves the opacity problem while allowing greatest consistent in the code/behavior

manfromjupyter (approved):
Oh, good stuff!!! (only I didn't say "stuff")

I love it! Looks great

PR #762, opened 2023-01-06
Remove pluck

### LINKED ISSUE #759: Phase out use of .pluck (deprecated)

mcrutch:
Closes #759 
Putting in as a draft PR so it is easier to track as each thing is tested in the review page of hte PR


app/models/review.rb
```diff
@@ -40,7 +40,7 @@ def recommended_reviewer?(user)
 
   # Is this review "recent"?
   def recent?
-    latest_revision_id = notebook.revisions.order(id: :desc).limit(1).pluck(:id).first
+    latest_revision_id = notebook.revisions.order(id: :desc).limit(1).map(&:id).first
```
mcrutch, line 43: This should probably just be .first.id rather than .map.first.  the limit(1) and .first might be redundant anyway, research this

PR #764, opened 2023-01-10
Making star and subscription icon consistent with others

### LINKED ISSUE #724: Ultra Dark theme make subscribe hover color same
On notebooks it's yellow but groups, tags, and users it's blue.

mcrutch:
I was thinking about this so I knocked out option 1
Closes #724


manfromjupyter:
@mcrutch, Confirmed that your changes have made it consistent across default and ultra dark theme across all pages.

Have some issue with dark theme (regular) so I can fix that but first:

Is there any way I can convince you to the number 2 solution to make stars and subscribes just consistent? Maybe it's fine but I just feel these features won't stand out if they're blue or orange next to the same and I feel like they should be treated as separate things because it's like a status thing where the others are all strictly actions. I recognize you've cleaned up the code a bit because it has gotten a bit too hacky for this but I believe I could carry on what you've done to clean it up and make it easy to locate in code with no hacked `!important`s. If you still decide you want solution 1, I respect it and will make those dark theme changes but just wanted to see if you might be open to the other first before making changes.

mcrutch:
I honestly hate it aesthetically when icons that are closely grouped and look the same have different colors on hover/action.  

Star, in addition to changing color, goes from outline to filled in in BOTH locations (the button itself an in nb_stats_view)
On the default theme Subscribe, amusingly enough, is almost the same color the RSS icon is usually displayed but regardless that will "highlighted"  on page load if they are subscribed without mousing over anything else which would still make it stand out IMO.

manfromjupyter:
@mccrutc, I understand. I'll just plan to keep all one color and just fix the dark theme before approving.

manfromjupyter:
@mcrutch, I understand. I'll just plan to keep all one color and just fix the dark theme before approving.

manfromjupyter:
@mcrutch, confirmed all colors work now. Also merge latest things from main into it (just ignore the double alerts appearing... Ima fix that now).

Issue #181, opened 2020-06-18
Ability to provide a user-friendly version # on upload of a new (and initial) revision (separate from the commit_id)


manfromjupyter:
Instead of version "9f4a4a33", it can be whatever they want as long as it's alphanumeric, below 8 characters, but including periods. Version "3.0C", "1.1.2", "8", "D", etc. 

Not sure if this is even possible but will look cleaner when using older revisions, doing version compares, etc and will make it easier to talk about in a business environment (ex. "Version 1.0 to 3.6B work as well as 4.1 to 5.5. 3.6C was giving people problems, but we fixed that pretty quickly.")

mcrutch:
It wouldn't be an instead of, it will be an also.  the version IDs that are there there are how they are tracked in git, so they have to remain.  The friendly display version would have to be tracked in addition to that.

manfromjupyter:
Oh right. Makes sense. I like it.

PR #768, opened 2023-01-13
Friendly Revision Labels

manfromjupyter:
Done:
- Ability to set, edit, and delete friendly labels on revision page for each label.
- Ability to set a friendly label (showing you the current one) on Upload New Version and on Approve Change Request forms.
- Made placeholder text consistent for optional fields
- Comparing versions uses the friendly label where it exists. Same for the banners and everything else.
- Made readonly fields consistent in code.
- Started a refactor on how we link things (to do things a way all frameworks accept and allow us more control, not just doing it the Rails way).
- Users can reset the version back to original UUID by setting an empty label. 
- Successfully sets all friendly label saves as inline alerts that doesn't reload page (to clean up one's versions super quickly).
- MAKE SURE that users still always have a way to still see the full original UUID.
- Fixed issue where revisions page with no comparisons to show was breaking datatable to fire a JS error (not rendering search for the table, not rendering pagination, and not added various classes to the table elements).
- Added error catch if trying to set version to one that already exists or setting one longer than 12 characters.
- Gave approver of change requests the decision of what version label to set it to (not the requestor)

Will do in another ticket:
- (To make consistent with label behavior) Saving/editing **revision summaries** send inline-alerts without it reloading page (so users can clean up summaries of all change logs of their notebooks super quickly).
- (To un-Ruby how we do links so it's more transferrable to other frameworks) Change how we do our `link_to_notebooks`, `link_to_groups`, and `link_to_users` to instead use the `_link.slim` way. This way also works in frameworks like Django Python so see it as a less-Ruby-only way. Will also give us more control over these links (HTML-wise, style-wise, and accessibility-wise).

Closes: #181


mcrutch:
Reviews table does not respect custom version numbers 
* notebooks/#####/reviews
* /reviews

mcrutch:
Do not do this:
> (To un-Ruby how we do links so it's more transferrable to other frameworks) Change how we do our link_to_notebooks, link_to_groups, and link_to_users to instead use the _link.slim way. This way also works in frameworks like Django Python so see it as a less-Ruby-only way. Will also give us more control over these links (HTML-wise, style-wise, and accessibility-wise).

Without ensuring you factor in what would be required for the #718 merge request.  Those magic helpers exist for a reason.

app/controllers/application_controller.rb
```diff
@@ -580,6 +580,26 @@ def verify_owner
...
+      return false
+    end
+    # Ensure new labels are shorter than 12 characters
+    if new_label.length > 12
+      return "Version label was too long. Only accepts a maximum of 12 characters and you submitted one that was #{new_label.length} characters. "
+    end
+    # Ensure new label is not one that already exists for that notebook
+    revisions = Revision.where(notebook_id: notebook.id)
+    revisions.each do |rev|
+      if rev.friendly_label == new_label
```
mcrutch, line 596: This throws an error if I accidentally click edit, and then immediately click save to not change the label.

app/controllers/revisions_controller.rb
```diff
@@ -57,12 +57,43 @@ def latest_diff
     end
   end
 
+  # PATCH /notebooks/:notebook_id/revisions/:commit_id/edit_friendly_label
+  def edit_friendly_label
+    errors = ""
+    notebook = Notebook.find(@revision.notebook_id)
+    friendly_label = params[:friendly_label].strip
+    if verify_revision_label(friendly_label, notebook, params[:old_label])
```
mcrutch, line 65: You are calling the same function twice with the same parameters.  Save the result the first time you call it and then append it to the errors. Don't call it twice.

app/controllers/revisions_controller.rb
```diff
@@ -57,12 +57,43 @@ def latest_diff
...
+    end
+  end
+
   # PATCH /notebooks/:notebook_id/revisions/:commit_id/edit_summary
   def edit_summary
     errors = ""
     revision_summary = params[:summary].strip
     if revision_summary.length > 250
-      errors += "Revision summary was too long. Only accepts 250 characters and you submitted one that was #{revision_summary.length} characters."
+      errors += "Revision summary was too long. Only accepts a maximum of 250 characters and you submitted one that was #{revision_summary.length} characters. "
```
mcrutch, line 96: "Only accepts...." This still feels like clumsy grammar even with the edit
"Summary can be a maximum of 250 characters..." ?

app/controllers/application_controller.rb
mcrutch: Ok, slightly different after I tried again. Click edit, change it, save it, click edit again, don't change it, click save, throws an error.

app/controllers/change_requests_controller.rb
```diff
@@ -154,9 +155,16 @@ def accept
       "#{@user.user_name}: [edit] #{@notebook.title}\n" \
       "Accepted change request from #{@change_request.requestor.user_name}"
 
+    # Revision Label Validation
+    if GalleryConfig.storage.track_revisions && params[:friendly_label] != ""
+      if verify_revision_label(params[:friendly_label], @notebook)
+        errors += verify_revision_label(params[:friendly_label], @notebook)
```
mcrutch, line 161: Another spot where you are double calling the same function.

mcrutch (changes requested):

manfromjupyter:
> Reviews table does not respect custom version numbers
> 
> * notebooks/#####/reviews
> * /reviews

Fixed this. Looked like the id to revision was only thing being used on review pages. Also fixed an issue where metadata Revision is technically the one being reviewed but that cannot receive a friendly_label.

app/controllers/change_requests_controller.rb
manfromjupyter: Thank you. Good catch. My mistake. Fixed.

app/controllers/revisions_controller.rb
manfromjupyter: Thank you. Good catch. My mistake. Fixed.

mcrutch:
Found one more annoyance.  If you have a long revision number it is a bit awkward because the only way to get the pencil with mouse is to mouse over the version number itself and then move right because there is insufficient right padding to allow the glyphicon to be there otherwise
image

app/helpers/application_helper.rb
```diff
@@ -79,10 +79,6 @@ def revision_path(rev)
     notebook_revision_path(rev.notebook, rev)
   end
 
-  def link_to_revision(rev)
```
mcrutch, line 82: For backwards compatibility, I think this should stay but at most throw a rails warning about deprecation.

manfromjupyter: That makes sense. Will do

app/controllers/revisions_controller.rb
manfromjupyter: Fixed

manfromjupyter:
> Found one more annoyance. If you have a long revision number it is a bit awkward because the only way to get the pencil with mouse is to mouse over the version number itself and then move right because there is insufficient right padding to allow the glyphicon to be there otherwise image

I was worried about the column getting too wide but I see it's not as big of an issue as I thought. Fixed!

app/controllers/application_controller.rb
manfromjupyter: Decided to do it through the front-end not via the backend so it doesn't even hit it (less times hitting our database).

mcrutch (approved):

app/helpers/application_helper.rb
```diff
@@ -80,6 +80,7 @@ def revision_path(rev)
   end
 
   def link_to_revision(rev)
+    logger.error('The rails helper "link_to_revision" is deprecated for this application. To ensure your revision links are able to show the user-friendly label, use the _link.slim partial instead.')
```
mcrutch, line 83: Not error, warn. and it must be called as Rails.logger.warn
`Rails.logger.warn("error message")`

manfromjupyter: @mcrutch, done. Sorry, got mixed up with Django for the logger part.

app/views/application/_reviews.slim
```diff
@@ -49,7 +49,7 @@
                   -if review.recent?
                     | Current
                   -else
-                    ==link_to_revision(review.revision)
```
mcrutch, line 52: is review.revision not already a revision object? (Same with below for row.revision)

app/views/revisions/index.slim
```diff
@@ -81,7 +92,30 @@ div.content-container
...
+                div.mouseoveredit
+                  ==render partial: 'link', locals: { revision: rev }
+                  span.sr-only #{" "}
+                  span.hidden aria-hidden="true" #{"["}
+                  a.edit-icon.tooltips href="#" title="Edit revision label"
+                    span.glyphicon.glyphicon-pencil aria-hidden="true"
+                    span.sr-only Edit revision label
+                  span.hidden aria-hidden="true" #{"]"}
+                form.revision-label-edit-form style="display: none" method="patch" data-url="#{revision_path(rev)}"
+                  -if rev.friendly_label != nil
```
mcrutch, line 105: Why do you need this differentiation. if rev.friendly_label is nil, it will just populate with empty values, no?  Just want to make sure we don't have an unnecessary if/then block that makes two almost identical lines sets of code for no reason.

manfromjupyter: My bad. Yep. You're right. For some reason I thought a `nil` value in a string would cause an error.

app/views/application/_reviews.slim
manfromjupyter: You're right. I had done this because I could not figure out why it was coming up as nothing because metadata and non-metadata revisions have the same commit_id. This has been fixed in the actual helper. Reverting this. Good catch.

PR #770, opened 2023-01-13
Solves issues with tags with " in them

### LINKED ISSUE #766: Tags with double quote causes error
Adding a quote of `hello"there` causes a super long 500 error, not the standard bag notebook error for the tag.

Potential Solution:
1. Don't physically allow the adding of quotes. Make it a deliminator of a new tag like we do with underscores, spaces, and some other characters.
2. Escape the character when it is added and then file the standard "bad notebook" error for the tag.

mcrutch:
1 - Makes it harder for that to occur by changing the field tokenizer 
2 - Makes the serve more gracefully handle errors from the CSV Parser in the future
Closes #766

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -704,7 +704,7 @@ $(document).ready(function(){
       success: function(json){
         $('#importTags').tokenfield({
           minLength:0,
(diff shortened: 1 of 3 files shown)


manfromjupyter (approved):
Flawless! No hard or even soft errors... utterly flawless.

PR #772, opened 2023-01-18
Make the full Forgot your Password form work properly with Ajax and a…

### LINKED ISSUE #719: Password reset confirmation just sends you to a JSON page
just says `{success: true}`

mcrutch:
…lerts

Closes #719

### DIFF
--- a/app/assets/javascripts/login.js
+++ b/app/assets/javascripts/login.js
@@ -46,4 +46,16 @@ $(document).ready(function() {
       $(".login-flashes").append("<p class='text-danger'>Please verify it is correct and try again</p>");
     })
   }
+  if(element=document.querySelector("#reset_password_full")){
+    element.addEventListener("ajax:success", function(event) {
+      console.log("Successfully reset password");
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/login.js
```diff
@@ -46,4 +46,16 @@ $(document).ready(function() {
       $(".login-flashes").append("<p class='text-danger'>Please verify it is correct and try again</p>");
```
manfromjupyter, line 46: For all of the text-danger and text-successes, could we add a role='alert' on these too just for accessibility stuff for the workflow of resetting password for example from the modal, not from the reset password page.

ex.
```js
$(".login-flashes").append("<p class='text-danger'>Invalid e-mail address</p>");
```
turn into:
```js
$(".login-flashes").append("<p class='text-danger' role='alert'>Invalid e-mail address</p>");
```

mcrutch: We should loop that in under another ticket since that is related to the other login/change password workflow from the pop-up dialog (versus from the /users/sign_in URL)

manfromjupyter (approved):

PR #773, opened 2023-01-18
Re-work Run in Jupyter dialog

### LINKED ISSUE #424: Run in Jupyter modal needs to be pulled out of JS code for Accessibility and styling fixes
Currently the modal is hardcoded. Proposal is to keep the scripting the same, but take out those hacks in place of a rails-made modal.

Need to consistency across the product, accessibility in the future, and one annoying bug for some people:
"long URLs wrap outside the table when multiple environments exists but none are default"

### LINKED ISSUE #591: The "Select Jupyter environment" modal needs to be generated the way we generate other modals
Currently it is generated via a bootbox-alert via JS (if I am seeing it right). What issues this causes (despite being styled the same) is for low visions users, when zoomed at 400% you can't scroll up or down to read the full dialog and for screenreaders and abulatory users, the focus is not trapped within the modal; you can tab your way out of it.

Also missing a modal description, etc

mcrutch:
Closes #424 
Closes #591


app/assets/javascripts/run_in_jupyter.js.erb
```diff
@@ -10,15 +10,28 @@ $(document).ready(function() {
...
+    var name =  $('#runInJupyterName').val();
+    var url =  $('#runInJupyterUrl').val();
+    var notebook = $('#runInJupyterNotebook').val();
+    var user_interface =  $('#runInJupyterUserInterface').val();
+    send(name,url,notebook,user_interface);
+    $('#runInJupyterOverwrite').modal('hide');
+  });
 
+  $(document).on('click','.run_jupyter', function() {
+    loadingGif();
```
mcrutch, line 34: This won't work on most pages yet.  Might roll back to the dialog for now, or just skip having a "opening" dialog at all, and then come back and do loadingGif here later maybe?

manfromjupyter (approved):
Code looks good and confirmed it works for failed environments and selecting which environment to run. Couldn't test live run or overwriting notebook dialog but it looks good.

PR #788, opened 2023-02-01
Allow homepage tabs to be bookmarked/linked to directly

### LINKED ISSUE #780: Allow direct linking to homepage tabs
Allow users to also save their preferred default tab?

mcrutch:
Does not affect forward/back browser activity
Closes #780

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -200,19 +200,17 @@ $(document).ready(function(){
   /* ======================================= */
   /* ======= Home Page Functionality ======= */
   /* ======================================= */
-  $(document).ready(function(){
+  function toggleActiveTab(element){
     var selector = '.page-home .tabs a'
-    $(selector).on('click', function(){
-      $(selector).closest("li.tab").removeClass('active');
-      $(selector).attr("aria-expanded", "false");
-      $(this).closest("li.tab").addClass('active');
-      $(this).attr("aria-expanded", "true");
-    });
-  });
-
+    $(selector).closest("li.tab").removeClass('active');
+    $(selector).attr("aria-expanded", "false");
+    element.closest("li.tab").addClass('active');
+    element.attr("aria-expanded", "true");
+  }
   currentTab='';
   function homepage_tab_load(tab = ''){
     currentTab=tab;
+    tabs = ["recent", "stars", "suggested", "mine", "all"];
     $("#txt").css("opacity", ".5");
     $("#newHomePageNotebooks").css("opacity", ".5");
     $("wait").css("display", "block");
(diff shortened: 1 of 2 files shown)


manfromjupyter (commented):
Confirmed it works, but think it's an improper syntax in the URL. "#" are anchor tags and if so should jump the user to that part of the page and start their taborder there. I can do that as is if I go to "https://SITE/#newHomeNotebooksRecommended" for example, but what you're doing here I think should be a param (ex. https://SITE/?suggested).

mcrutch:
Location hash routing has become a [proper design pattern](https://angular.io/api/common/HashLocationStrategy) since the advent of JavaScript heavy frameworks/UIS as changing a get parameter on a URL requires navigating to that URL. (Eg I can change location.hash without causing the webserver to want to reload the page, but I can't change a get parameter without triggering the browser navigating to a new page as that is a distinct server-side URI).  This action is 100% controlled by JavaScript, not by the server, so you want to use something that is simple to parse in Javascript.

By changing the location hash when the user also clicks the link, it means that the URL in their browser bar is ALSO the url for that tab that they are currently viewing so a simple copy paste of the URL gets them to that current homepage tab.  Parsing out the get parameters from a URL in JavaScript is more complicated than just getting the location hash because other parameters could for one reason or another end up tagged onto that URL at any point.

manfromjupyter:
@mcrutch I understand the use case of not wanting to reload the page on tab change but also wanting to gather a helpful URL. 1. Would we be open to adding those ids to the sections so they also act like anchors?
example solution:
```html
        ul.tabs
          -if @user.member?
            li.tab id="suggested"
              a.tabLink href="#suggested" aria-label="Show notebooks recommended for me" aria-expanded="false" id="newHomeNotebooksRecommended" Recommended
          li.tab id="all"
            a.tabLink href="#all" aria-label="Show all notebooks" aria-expanded="false" id="newHomeNotebooksAll" All Notebooks
          li.tab id="recent"
            a.tabLink href="#recent" aria-label="Show notebooks recently viewed" aria-expanded="false" id="newHomeNotebooksRecent" Recent Notebooks
          -if @user.member?
            li.tab id="mine"
              a.tabLink href="#mine" aria-label="Show my notebooks" aria-expanded="false" id="newHomeNotebooksYours" My Notebooks
            li.tab id="stars"
              a.tabLink href="#stars" aria-label="Show all of my stars notebooks" aria-expanded="false" id="newHomeNotebooksStars" Starred Notebooks
```

2. Are we open to keeping the recommended field set to no param/href? We keep it without an anchor or href because its the default view? That way we can keep the URL a little cleaner sometimes?

mcrutch:
Recommended is only the default view for logged in users.  The default tab changes based on if the user is logged in or not and eventually by user preference too (longer-term plan).  

Back/Forward are ignoring the location hash since I am using history replace. 

Adding the ids to the divs might cause a screen jump every time they click on a tab which would be undesirable if it does.

manfromjupyter:
Makes sense. Approved

Issue #784, opened 2023-01-25
cleanup un-used gems


mcrutch:
start with coffeescript

PR #789, opened 2023-02-01
Droplet

mcrutch:
Related to #784


manfromjupyter (approved):
@mcrutch code looks good. Need a full test too?

mcrutch:
We should probably do a speed run through the site. I took out a lot of gems that I don't think touched anything relevant and didn't see reference to these gems else ware but there is a SLIGHT chance something was used that wasn't properly listed as a primary dependency rather than just a transitive dependency

mcrutch:
Took a different path, went with harder verification but made sure these gems are used anymore, gonna merge.

PR #792, opened 2023-02-02
Uses makeAlert for the login modal errors

### LINKED ISSUE #775: Fix login alerts/errors for login modal
https://github.com/nbgallery/nbgallery/pull/772#discussion_r1081721694

mcrutch:
Closes #775

### DIFF
--- a/app/assets/javascripts/common.js
+++ b/app/assets/javascripts/common.js
@@ -6,6 +6,7 @@ function makeAlert(type, element_within, message){
   if (element_within == ''){
     element_within = '#main .alert-container:first';
   }
+  $(element_within).children().remove();
   // NOTE - this shall only be used to generate messages on pages that are not going to be reloaded. If you wish to reload the page AND have an alert when it loads, alert needs to be generated from backend.
   if (type == "success"){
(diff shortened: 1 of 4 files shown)


manfromjupyter (approved):
@mcrutch, Looks good and works! Only things I'd recommend are: (if you agree)
1. Get rid of the newline on the makeAlert on line 39 of app/asets/javascripts/login.js because it makes it looks weird:
same line success message when registering

2. Something I forgot a long time ago when I added that `makeAlert()` is to clear all other alerts when we add a new inline alert (for that given container). Example:
think we might need to have new inline alerts clear all others in the container

Should edit `makeAlert()` method and inject the following between `line 8 and line 9`: https://github.com/nbgallery/nbgallery/blob/main/app/assets/javascripts/common.js#L9
**Code to add:**
```js
$(element_within).children().remove();
```

mcrutch:
I thought we couldn't remove errors until dismissed by the user for accessibility reasons?

manfromjupyter:
This is ready now @mcrutch when you're ready. I did the following:
1. Fixed the multiple-alerts issue where they just stack forever.
2. Cleaned up what login messages where there.
3. Fixed the padding of alerts in the Login Modal.
4. Added several more use-cases for firing alerts on the Login Modal.
5. FINALLY fixed the multiple `id` issue Chrome keeps firing on Inspect Element > Console. Had to set the ids to `nil` in Ruby to not render. Just deleting them or setting as nothing wasn't enough.

PR #795, opened 2023-02-02
Updating gems

### LINKED ISSUE #726: investigate upgrading to oauth2 2.x+ gem

mcrutch:
Closes #726


mcrutch:
This updates a lot of gems.  There are more that COULD be updated but they definitely break things.  
execjs >= 2.8 would require replacing therubyracer with mini_racer or another option
browser has breaking changes
there might have been one or two others as well.

manfromjupyter (approved):
Tested everything I could think. All looks good.

PR #811, opened 2023-02-16
Group Error Message/Backend Refactor

### LINKED ISSUE #806: Group Edits with missing fields causes weird error
Updating groups causes weird error if missing fields

manfromjupyter:
Closes #806 

And addresses the concerns of #812. We'll always have the cosmetic issue but at least it will catch it and explain what is needed every time if user clicks the grayed out button anyway (or it isn't graying out)

1. Refactored back-end of groups to follow the format of our other controllers.
2. Now appropriately handle all group errors I can think of as soft errors allowing the user to fix them without starting over.
3. All errors now mention what went wrong and how to fix them (an accessibility requirement).


mcrutch:
It would be better if we gave a more meaningful error from the server side rather than just list all the possible reasons it failed in Javascript.

manfromjupyter:
NOTE TO SELF - just need to revert my deletions to member_list method and this is ready to go

manfromjupyter:
@mcrutch Merge conflicts fixed for this now

PR #814, opened 2023-02-23
Fixed wrapping issues in admin and user detail's tables

### LINKED ISSUE #793: User Details Similar Users table may extend off page if a user has long email
Make emails wrap on that table.

manfromjupyter:
- Fixed wrapping issue with users, groups, tags, and notebooks on all of the admin pages with graphs and the user details page.
- Fixed table-compression issue of users with too long of an email or username on the /users page.

Closes #793

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -1231,10 +1231,10 @@ $(document).ready(function(){
     'searching': false,
     'info': false,
     'columnDefs': [
-      { 'responsivePriority': 1000, 'targets':[4,5] },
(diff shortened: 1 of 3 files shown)


mcrutch:
Compare to username on our dev site.  400% zoom fails (adds horizontal scroll) still

manfromjupyter:
@mcrutch Fixed. Appears datatables was failing due to the table with email column causing 1 fewer column than expected. 

If you are still seeing the issue; try refreshing your page. It will re-adjust the max table width.

mcrutch:
Also fixes #818

PR #819, opened 2023-02-27
Remove bootbox pending

mcrutch:
Doesn't work for the dialog autocomplete stuff right now (like changing owners)


mcrutch:
Ok, This is the next to last item in death-to-boot box (last item will be two confirmation dialogs).
Take a look at how this works and let me know what you think @manfromjupyter

manfromjupyter (approved):
Code looks good and the solution is elegant as hell.

The only problems/changes I will do in another ticket or aren't worth fixing.
1. loading gif on dark theme on Users page is white on white background but I'll fix in another ticket.
2. I'd like the base loading gif to be bigger as you really don't even see it some of the time.
3. Loading gif should be showing up on some pages but our page loads are so good now that it sometimes only shows up for a millisecond and a few times not at all but I don't think it's worth taking it out of the code for those given endpoints.

PR #821, opened 2023-02-27
Fix deleting stages from the stage listing screen

### LINKED ISSUE #820: deleteStagedNotebook id reused but assumed to be signular
All three delete icons use the same element ID and only the third actually triggers the javascript
image

mcrutch:
Also remove one of the 2 last bootbox confirmation dialogs
Closes #820


manfromjupyter (approved):
Beautiful and flawless. Confirmed worked on both the index and show pages and returns to right back on deletion.

PR #824, opened 2023-02-28
Check for .blank? on revision.commit_message

### LINKED ISSUE #670: Saving an Empty Commit on Revision History prevents you from ever editing it again
If you delete all of the characters of a commit message on Revision History, you can't ever edit it again to add a description... BU if there was no commit message to begin with, you can edit it (unless you just immediately save it while it has no characters.. then you're in the same boat.

mcrutch:
nil is blank
"" is blank
Closes #670

### DIFF
--- a/app/views/revisions/index.slim
+++ b/app/views/revisions/index.slim
@@ -137,7 +137,7 @@ div.content-container
             -else
               td.commit-message-cell
                 div.mouseoveredit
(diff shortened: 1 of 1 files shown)


manfromjupyter (approved):
Looks good and confirmed it works.. Wow that's all that was needed. I don't know why I kept putting it off.

PR #828, opened 2023-03-14
Display group description on Notebook listing

### LINKED ISSUE #301: Expose Group description on group pages most likely on the banner
May take some finessing and edge case testing as real estate is limited

mcrutch:
Closes #301 
@manfromjupyter You are the UI man, but here is what I am thinking.  It only shows up on the listing, not with the landing page as a landing page should be self-descriptive.


manfromjupyter (approved):
Hovering over the group names on the /groups page over the ones with markdown looks weird, but love the functionality and can prove it works and code looks good.

PR #829, opened 2023-03-14
Fork awesome

### LINKED ISSUE #765: Worth/possible to only use one of glyphicon/font acid icons?
We have both scattered all over the place.  I don't think glyphicons are complete enough for our purposes, but could fa icons be?

mcrutch:
Closes #765 
Removes Font-Awesome
Adds Fork-Awesome
Replaces almost all Glyphicon usage with Fork Awesome
(Only remaining Glyphicon usage is by Bootstrap Form Validator)


manfromjupyter (approved):
Checked every page and every icon. I am satisfied with it all. I did not go back to improve our "enlarge font-size" preference but it not breaking, just a couple of places that could be made larger. I am not sure users are even using this or we want to take it on though (in favor of just making people zoom in if they want to make things bigger)... the juice might not be worth the squeeze but we'll sync later on that.

PR #668, opened 2022-06-05
Death to bootbox alerts PART 1

manfromjupyter:

Issue #590, opened 2021-08-18
Low Vision issue when encountering bootbox alert issues while on modals


manfromjupyter:
Zooming in to 400%, scroll down on modal (ex. /environments), encounter bootbox alert (ex. try submitting it with a an environment name that contains a space character), and closing said alert won't let you scroll back up.

Should use our custom alerts instead. No additional popups.

mcrutch:
Bootbox is dead

PR #852, opened 2023-05-18
External Resource Deletion -Last bootbox alert

### LINKED ISSUE #182: Replace all bootstrap confirmation dialogs with a singular one we can customize accessibility wise
Replace all bootstrap confirmation dialogs with a singular one we can customize accessibility wise

- Coloring of buttons is not consistent with our product
- Tabbing through the modal is not consistent with our product or fully accessible
- Missing modal header
- Verbiage can just be "Are you sure?" and will probably be a second quicker than the other dialog because it's loaded with the page DOM. This new dialog can apply to all places we already have one (ex. deleting notebook) and add one where we need them but don't at the moment (ex. deleting environment, deleting org chart, etc)

manfromjupyter:
Closes #182 and #590 

1. Made delete external resource dialog follow same format as rest of our product.
2. Final death to bootbox #668.
3. Moved delete external resource JS to an actual JS file to be minified.
4. Made it so viewing delete notebook confirmation and then going to delete resource doesn't cause those addition line items text to appear.

After this we can do out best to look into what extensions people have with the product and can remove the require bootbox.min and additional bootbox CSS if it comes up clear. @mcrutch if you think we're good to do this now though, just let me know.


app/assets/javascripts/notebook_actions.js.erb
```diff
@@ -218,6 +218,63 @@ $(document).ready(function(){
...
   });
 
+  /* ===== Delete External Resource ===== */
+  $('.external-resources .delete-icon').on('click',function(e){
+    e.preventDefault();
+    $('#confirmationModal').modal('show');
+    $('#confirmationModalForm').addClass('delete-resource');
+    var text = 'Delete the resource: "' + $(this).attr('data-resource-title') +  '" from notebook: "' + $(this).attr('data-notebook') + '".'
+    $('#confirmation1').text(text);
+    $('#confirmationModalForm #inAdditionText').addClass('hide').siblings('.additional-confirmation').text('').addClass('hide');
```
mcrutch, line 228: Click to delete external resource
Click Cancel
Click to delete Notebook
Additional Info list is missing in dialog

mcrutch (changes requested):
See comment about Additional Info

manfromjupyter:
@mcrutch Fixed both no matter what one confirmation is selected first and made sure to clear the text completely from one to the other.

mcrutch (approved):

Issue #253, opened 2020-07-28
Request for Azure AD oauth


mcrutch:
See comments of #249 
Need to make sure gems all work nicely, but also don't have an environment to test that in.
https://github.com/marknadig/omniauth-azure-oauth2

cccs-cat001:
So.. is no one interested in doing this? :P

manfromjupyter:
Hi @cccs-cat001, I know we currently support authentication via Gitlab, Github, Facebook, and Google but because none of us have an Azure account, we haven't done it because we can't test it and we didn't want to release anything we couldn't test. That is my best understanding. Perhaps it's because the gem is no longer maintained but I just don't know. We have a couple people out at the moment. We can look at this in a couple months. I'm sure there is interest just not sure where we left off with this one or what the problems where if there were any.

cccs-cat001:
I'm not a ruby dev by any means.. but if someone can guide me through the setup I'd be willing to try it out and test it, I have an azure account

manfromjupyter:
@cccs-cat001 I unfortunately don't think we have the human power/bandwidth at the moment to give you the amount of detail we'd hope we could give you on this but I can occasionally provide responses but I have no experience with doing this exact thing. In theory look for every mention of `"facebook"` in the code, make a copy of it, and replace with `"azure"`. Ex. instead of adding another `omniauth-facebook` gem, it would be `omniauth-azure-oauth2`.  They have some directions here: https://github.com/marknadig/omniauth-azure-oauth2. Add the gem to your gemfile. Gems are essentially "packages" as you would see for python for example, and then the `Gemfile` and `Gemfile.lock` are just the requirements and install list. In the config I think there is a place for enabling or disabling certain types of authentication... May just need to enable to disable that. As per the rest I do not know, but can look up certain pieces of it if you get stuck. Worst case scenario we can reattempt this some time later once we get that key staff back in and will just be sure to sync with you to test it for us. I can't make any promises or decisions of what we can merge or work on, but these are just the assumptions I'm working on at the moment, apologies.

cccs-cat001:
Alright cool, the theory sounds easy lol I'll see what I can do :)

mcrutch:
Maybe @ScratZa has some documentation he can share for enabling the Azure oauth since he added the gem during his 5.2 upgrade mege request.

ScratZa:
hey chaps - apologies been  a while :) 
Have created a PR  with some changes ive tested my side - seems to be working as expected.

PR #856, opened 2023-05-19
Auth/add ad login

ScratZa:
Added Azure AD as an option for Login.


ScratZa:
Resolves #253

mcrutch:
What is that "CACHED" file for?

ScratZa:
Removed the Cached file and also removed the azure settings from the dev docker script keeping it consistent with how it was previously.

Issue #730, opened 2022-12-22
Verified notebooks lose verification (approved reviews) if you make public/private


manfromjupyter:
If you make a notebook verified with reviews but then make it private or make it private and then public again, the verification is lost. Behind the scenes, making it private/public adds another revision (metadata) change and verification only looks are current version.. Either reviews need to not include metadata revisions OR making public/private shouldn't add a revision event.

Issue #578, opened 2021-08-02
Carry forward Queued Reviews on notebook updates


manfromjupyter:
I thought we caught this but appears now. This may be cleaned up automatically as part of our nightly cron job (because I could have sworn we caught this but I don't know for certain and/or starting to to doubt it). Whenever you have queued reviews and then update the version, we are supposed to delete the reviews that had never been touched. If it was claimed or completed, it's useful knowing for historical purposes what versions were working/looked properly. As it relates to Queued notbeooks though, these reviews were never even assigned/claimed, let alone looked at.

Perhaps the proposed solution is to carry queued reviews forward if they aren't reviewed on an older version. Not sure why, for example, a notebook a week old would need a review but updating a few days after would NOT need a review if it was suddenly updated. I would assume the need for a review would still be needed; we might as well save the users from refilling out the form.

jeffyjefflabs:
Yeah, I agree reviews in "queued" status should always point to the latest revision until claimed.  Maybe if status is queued, then revision should just be null?  Then when claimed, set revision to the latest.  Whatever is easiest, I don't think the implementation really matters, we just don't want things to still be in the queue if there's already a newer revision.

If I *complete* a review and then the author updates the notebook, then we should keep the review tied to the older revision, since that's the code I looked at.

If I *claim* a review and then the author updates the notebook before I finish the review, that's kind of tricky and there's probably not a perfect solution.  Maybe something like an email notification to me (the reviewer) that the notebook has been updated, and then I get an option (or am forced?) to update to the latest revision?

mcrutch:
So, since we were talking about having all reviews done possibly enabling other things....what do you do if one review is completed, two are queued, and the notebook updates?  Do you want to try and get all three reviews done on the same revision or do you pull forward the two queued?

jeffyjefflabs:
Hmm that's another tough one, because what if it was a major change, or there have been multiple changes, or the completed review is 6 months old and the queued ones have been sitting there the whole time.  At some point the completed review becomes "historical" and not relevant to current decisions about the notebook.

Maybe the different types of review should be linked together i.e. one super-review (pinned to a revision) that has 3 sub-reviews?  My concern there is that some types of review may not need to be done as frequently, so then does it look weird to have all these partially-completed (super-)reviews?

[...]

Issue #190, opened 2020-06-18
Review workflow Recommended Reviewer Add/Remove


manfromjupyter:
Review workflow Recommended Reviewer table needs way for users to add users to the table as well as give users the ability to remove themselves from the table. This is the last piece of one or two for the Review workflow from making it widely usable.

manfromjupyter:
Recommend doing after #857 in terms of the Review feature story.

PR #857, opened 2023-05-30
Individual Review History - tracking comments, reviewer, and status changes over time

### LINKED ISSUE #204: Add comments input field to every step of review workflow
Furthermore, inject pre-generated text in these so users know this is changing and know it won't be carried over between interactions. Ex. if user Jim claims review, comment by default should say "Jim claimed Functional Review... if they want to add more they can, but they know before they approve, etc what will be commented.

manfromjupyter:
- Closes #204 and is 3rd to last thing before getting reviews fully functional.
- lets screen readers know when certain reviews are missing/were not proposed on all Reviews page.

## Review Roadmap IMO
1. **This PR:**
    - have page or modal users can visit that shows the comments all steps of the way for an individual review, what the status was for each and when as well as all of the changes to the reviewer if there were any.
    - allow comments to be placed every step of the process.
2. **Next PR:** #190 
    - ability for reviewers to opt out of the top 10 if a user was placed on there by another viewer or the system automatically
    - ability to add users manually
    - ability to remove users manually
3. **Eventually:**
    - allow any notebook owner to nominate their notebook for review.
    - turn on numeric ping in the header like change requests when a user is appearing as a recommended reviewer of a notebook.
    - email notification when you the user is manually added to a review, else as part of the subscription email at the start of every day if newly added or when any status changes for any review on notebooks they own (i.e. review is claimed and comment is left)
    - OPTIONAL - feature within app or link to a checklist per type of review what the reviewer needs to review against
    - OPTIONAL (#730) - feature to not lose verified status if it's been reviewed
    - OPTIONAL (#578) - queued reviews stay queued if notebook is updated

If we want to set up ability to turn on allow-lists for certain types of reviews and allow admins to maintain this allow-list for these types of allow-lists if for example compliance reviews can only be done by certain individuals, then we can do this after 3?


manfromjupyter:
BLOCKED on this until I can get a `rail generate model` to work with our Github Ci/CD runners or turn it off (which seems to be what's blocking my command or manual creation of it from working)

manfromjupyter:
Just noting am no longer blocked but going to focus on the other things now.
```
MySQL [gallery]> show columns from review_history;
```
| Field       | Type         | Null | Key | Default | Extra          |
| ------------- | ------------- |  ------------- | ------------- | ------------- | ------------- |
| id          | int(11)      | NO   | PRI | NULL    | auto_increment |
| review_id   | int(11)      | NO   | MUL | NULL    |                |
| user_id     | int(11)      | YES  | MUL | NULL    |                |
| action      | varchar(191) | NO   |     | NULL    |                |
| reviewer_id | int(11)      | YES  |     | NULL    |                |
| comment     | text         | YES  |     | NULL    |                |

manfromjupyter:
**Left to do on this PR**
1. More links back to forth to the review history
2. Ability to view review histories of all reviews from one page OR each review can link to the history OR both?
3. Claim, Unclaim, and Approve all prompt modals to add a comment if the person wishes.

manfromjupyter:
@mcrutch to test this on any new database, just need to do a fresh DB migrate and then set config for all 3 types of reviews to in the `settings.yml` to "true".

mcrutch:
I feel like maybe this should say "approved" rather than completed even though the only way to get to completed is approved, I just feel like it would be more clear. Thoughts?
image
image

manfromjupyter:
> I feel like maybe this should say "approved" rather than completed even though the only way to get to completed is approved, I just feel like it would be more clear. Thoughts?

@mcrutch I've changed all instances of `completed` to `approved` in our database using a db migrate so other people will also get when they update to latest. I've also changed all logic in our codebase to look for "approved" not "completed" for those review or review_history statuses. Should be ready now.

app/views/modals/_review_action.slim
```diff
@@ -0,0 +1,18 @@
+div.modal.fade id="reviewActionModal" aria-labelledby="reviewActionHeader" aria-describedby="reviewActionDescription" role="dialog" style="display: none" tabindex="0"
+  div.modal-dialog
+    div.modal-content
+      div.modal-header
+        h1.modal-title id="reviewActionHeader" Review
+        p.sr-only id="reviewActionDescription" Dialog for adding a comment on claiming, unclaiming, or completing a reviee
```
mcrutch, line 6: "reviee"

PR #862, opened 2023-06-14
Setup custom email headers

### LINKED ISSUE #860: Allow plugin-like or single line (either) support for custom mail headers
Each call to mail will need to support sending extra headers.
Options:
add default headers to ApplicationMailer base class that pulls from GalleryConfig.email.custom_headers or something of the like
Have a function from ApplicationMailer base class that returns headers (that can call a plugin)
either way, every call to mail has to support it being called just before
Example: `headers['X-Special-Domain-Specific-Header'] = "SecretValue"`
Src: https://edgeapi.rubyonrails.org/classes/ActionMailer/Base.html#method-i-headers

manfromjupyter:
Closes #860 

@mcrutch is this what you were thinking?

added as header

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -77,9 +77,9 @@ email:
   email_headers:
     enabled: false
     headers:
-      - name: Example Header
-        key: X-Special-Domain-Specific-Header
-        value: SecretValue
+      #- name: Example Header
+      #  key: X-Special-Domain-Specific-Header
+      #  value: SecretValue
 
 # Instrumentation options
 instrumentation:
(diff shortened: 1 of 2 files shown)


config/settings.yml
```diff
@@ -74,6 +74,12 @@ email:
   general_from:
   exceptions_from:
   exceptions_to:
+  email_headers:
+    enabled: false
+    headers:
```
mcrutch, line 79: The lines defining the headers should be commented out to show what they would look like but not shove in unnecessary headers into the config

manfromjupyter: Okie doke. Thought that since they were set to false by default it would be okay, but that makes sense; smarter and safer

PR #863, opened 2023-06-16
Remove Bootbox Libraries entirely

mcrutch:


manfromjupyter (approved):
Looks good and don't see a potential issue with it. Tested a few dialogs anyway.

PR #894, opened 2023-08-09
[config, db] Remove dependency on the database during startup

hosamaly:
Attempting to load the application used to require a valid connection to the database, which prevented the use of commands like `rails db:create`.

This was caused by a statement in the `User` model class that attempted to connect to the database to find out the names of the fields in the `UserSummary` class in order to define delegator methods for each of them. A similar statement in the `Notebook` class had the same effect.

This change fixes the issue by allowing this delegation to happen later during the application's runtime.

In the Notebook class, the delegated methods have been hard-coded instead of being dynamically looked up. This is because some of them used a prefix while others didn't. Thus, new attributes will note be delegated by default.

The User class was easier because it treated all delegated methods the same, so it still allows for dynamic lookup. It also ensures that any methods defined on the User class are not delegated. As a side effect, though, it may delegate methods that are not attributes, but this seems like a small price to pay.

With this change, we can now run `rails db:create` and `db:test:prepare`. We can also run `rails console` even if we don't have access to the database.

Additionally, I added the database schema to source control. It is strongly recommended to track database schema changes, showing how these changes relate to the history of the project.

With the schema in the repository, we can now run `rails db:setup`, `rails db:schema:load`, and `rails db:test:prepare` out of the box.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,11 +9,11 @@ class User < ApplicationRecord
   }
   before_destroy { |user| Commontator::Comment.where(creator: user.id).destroy_all }
   before_destroy { |user| Subscription.where(sub_type: "user").where(sub_id: user.id).destroy_all }
-  # Include default devise modules. Others available are:
-  # :confirmable, :lockable, :timeoutable and :omniauthable
+
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable,
          :confirmable, :omniauthable, :timeoutable, :lockable
+
   has_one :preference, dependent: :destroy
   has_one :user_summary, dependent: :destroy, autosave: true
   has_one :user_preference, dependent: :destroy
@@ -130,8 +130,6 @@ class User < ApplicationRecord
   scope :filter_by_org, -> (org) { where( "org like ?","%#{org}%" ) }
   scope :filter_by_approved, -> (approved) { where( "approved = ? or (approved IS NULL and 0 = ?)", approved, approved ) }
   scope :filter_by_admin, -> (admin) { where admin: admin }
-
-  extend Forwardable
 
   def email_in_allowed_domain
     allowed_email_address = false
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -356,10 +356,7 @@ def change_requests_owned
   #########################################################
 
   # Delegate methods to summary object
-  UserSummary.attribute_names.each do |name|
-    next if name == 'id' || name.end_with?('_id', '_at')
-    def_delegator :user_summary, name.to_sym, name.to_sym
-  end
+  delegate_missing_to :user_summary
```
hosamaly, line 359: The call to `UserSummary.attribute_names` forced Rails to connect to the database to figure out the table structure. However, since this was happening in the class body, it got run as soon as the class was loaded, which takes place when `devise` is configured in the routes file.

This was relatively early and it didn't give the application a chance to finish initialization or run `after_initialize` hooks. It made the database connection a mandatory dependency to load the application, causing all Rake tasks to be unusable unless the connection can be established.

My suggested solution is to use `delegate_missing_to`, which avoids the dependency during startup and is more flexible with attribute-like methods (e.g. predicates such as `user_rep_raw?`).

db/schema.rb
```diff
@@ -0,0 +1,608 @@
...
+# of editing this file, please use the migrations feature of Active Record to
+# incrementally modify your database, and then regenerate this schema definition.
+#
+# This file is the source Rails uses to define your schema when running `bin/rails
+# db:schema:load`. When creating a new database, `bin/rails db:schema:load` tends to
+# be faster and is potentially less error prone than running all of your
+# migrations from scratch. Old migrations may fail to apply correctly if those
+# migrations use external dependencies or application code.
+#
+# It's strongly recommended that you check this file into your version control system.
```
hosamaly, line 11: Having this file in the git repository makes it much easier for anyone to learn about the data model of this project without requiring access to a running database.

hosamaly:
Sadly, these changes may not be enough. There is a similar issue in the `Notebook` class, but it's more complicated by the fact that some methods are given a prefix. :(

https://github.com/nbgallery/nbgallery/blob/a569ce56e9701e2419320c1f09d7011426fe4bce/app/models/notebook.rb#L736-L744

PR #899, opened 2023-08-29
make My Groups visible even if user has no groups

BrendanPC:
While looking the Groups bug, it occurred to me that there isn't really any way for users to create groups if they aren't already in one. This is a simple change that makes the groups functionality a bit more discoverable.

image

### DIFF
--- a/app/views/layouts/layout.slim
+++ b/app/views/layouts/layout.slim
@@ -225,16 +225,15 @@ html lang="en"
                             li
                               a href="#{subscriptions_path}"
                                 span.tab My Subscriptions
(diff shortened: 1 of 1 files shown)


mcrutch:
It is possible, just here:
image

BrendanPC:
> It is possible, just here: image

Whoops, somehow I missed that.

mcrutch:
So, I am not saying I am NOT going to merge, this, just pointing out there is a way.  I want to have a pro/con conversation with @manfromjupyter before I make my final call.

PR #907, opened 2023-09-08
Removed Jupyter Classic as an interface option

### LINKED ISSUE #906: Remove nbclassic from Environments launch option
Jupyter Classic has been deprecated by the opensource community so to prevent 404 errors need of people launching it, we should remove it as an option.

manfromjupyter:
Closes #906 

Removes the option of "Jupyter Classic" people may set for their environment and removes existing ones that have been set.


mcrutch:
That is a valid UI for Notebookv7 environments

Issue #848, opened 2023-05-15
Disable all Form submittals on disabled submit button submissions


manfromjupyter:
STOP accepting Enter key or Submit button clicks while the button is grayed out.

In the JS require for each method require the button not be disabled.

manfromjupyter:
Relates to #639 and readdresses #812 better

TODO as of #851 
1. Get all forms to disable the submit button by default if any required fields are not filled out.
2. Get all forms ideally to follow `disabled="true"`/`disabled="disabled` not just tack on `class"disabled"` because the latter is purely cosmetic.
3. Ensure submit buttons are `disabled="true"`/`disabled="disabled` for all forms WHILE form is processing/submitting so as to not allow any multiple-submittals.
4. Full audit that all forms needing `required="required"` get them

## Status as of #851 
| Form | Summary | Disable/Enable support status |
| ------------- | ------------- |  ------------- |
| Admin - Site Warning#Create  | Good (now) | Design only (but now disabled by default with this PR) |
| Admin - Site Warning#Edit  | Good (now) | **N/A** - will be all filled out already by default, just is disabled upon removing fields by Design only. |
| Admin - Site Warning#Delete | Perfect (now) | **DONE** - always able to be clicked after site warning has been created. Hidden on new site warnings (now) |
| Admin - Site Warning#Clear (NEW) | Perfect (new) | **DONE** - always able to be clicked on new site warning. |
| Admin - Import Notebooks  | Bad | Design only (and not disabled by default) |
| User (NBGallery) Preferences  | Perfect | **N/A** - always filled out  |
| Change Ownership  | Perfect | **DONE** (before) |
| Deprecate Notebook  | Perfect | **DONE** (before) |
| Share Notebook with Users | Perfect (now) | **DONE** (with this PR)  |
| Add Environment  | Bad | Disabled via design only |
| Edit Environment  | Bad | Disabled vis design only  |
| Propose for Review  | Bad | Design only (but not disabled by default) |
| Upload Notebook/Upload new Version - first modal  | Bad | Design only (and not disabled by default) |
| Upload Notebook - second modal  | Bad | Design only (and not disabled by default) |
| Upload new Notebook Version - second modal  | Bad | Design only (and not disabled by default) |
| Propose Change Request  - first modal  | Bad | Design only (and not disabled by default) |
| Propose Change Request - second modal  | Bad | Design only (and not disabled by default) |
[...]

PR #914, opened 2023-09-08
Fix Propose Review, better fix for external resources, working on env…

mcrutch:
…ironmnts

Relates to #848


mcrutch:
Closes out almost all of the remaining issues with #848.

Spend some time reviewing this one.

app/views/users/summary.slim
```diff
@@ -58,15 +49,15 @@ div.content-container
...
           div.input-group
             label.input-group-addon.input-min-width for="summaryTimeEnd" End Date
-            input.form-control.tooltips id="summaryTimeEnd" type="date" name="max_date" placeholder="2099-12-31" title="End Date" required="required" value="#{params[:max_date]}"
+            input.form-control.tooltips id="summaryTimeEnd" type="date" name="max_date" title="End Date" value="#{params[:max_date]}"
           span.glyphicon.form-control-feedback aria-hidden="true"
         div.button-container
-          a tabindex="-1"
+          a href="#{summary_user_path(@viewed_user)}" tabindex="-1"
+            button.btn.btn-primary type="reset" Reset
+          a id="filterUserSummarySubmit" tabindex="-1"
```
manfromjupyter, line 62: So resetting does work but not if you have submitted before.

Submitting changes the URL and sets the values equal to what was preloaded into the URL. Resetting only clears new values, not the values that were present at the initial state of the form (which was after last submit). You can clear before you hit submit but not after if you expect the whole thing to clear.

Think we need to ditch the native browser functionality and bring back the JS we had.

manfromjupyter:
**User Summary form** - `can't intuitively clear fields`
**Environments (Add/Edit/Delete)** - works perfectly
**NBGallery Preferences Form** - previously and still works
**Search Form** - still works as expected 
**User Forms (/users/ID/edit, /users/edit, etc.)** - still works as expected 
**Site Warning (Add/Edit/Delete)** - works perfectly
**Admin Import Notebooks** - still works as expected 
**Upload New Notebook (both modals)** - works perfectly
**Upload New Version (both modals)** - works perfectly
**Provide Feedback** - `design only still (forgot on the original ticket, #848 woops)`
**Propose Change Request (first modal)** - `design only still`
**Propose Change Request (second modal)** - works perfectly
**Deprecate Notebook (Add/Edit/Delete)** - works perfectly
**Propose for Review** - works perfectly
**Group (Add/Edit)** - `EXISTING ISSUE - adding new empty row for user lets user submit`. Do we want to fix and keep that a backend error?
**Approve/Cancel/Decline Change Requests** - works perfectly
**Share notebook** - works perfectly
**Change Owner** - works perfectly
**Notebook Title** - `EXISTING ISSUE - Can click submit button when it's empty` (forgot on the original ticket, #848 woops)
**Notebook Description** - `EXISTING ISSUE - Can click submit button when it's empty` (forgot on the original ticket, #848 woops)

manfromjupyter (approved):

PR #915, opened 2023-09-08
Modal Announcements for screen readers

### LINKED ISSUE #191: Modals need to be announced when opened (Accessibility)
This is super important if not required for accessibility. We have it so the focus changes and the next read or tab is in the modal (and the action leading up to the modals lets them know an action would happen), but an announcement is required.

manfromjupyter:
Closes #191 
- Fixed makeScreeenreader alert method so it works with the Tooltips one we made earlier too.
- Set it up to fire on all modals except those with have a need to return false and not trigger other methods but added screen reader alerts to those individually (to not break anything)
- Added aria-haspopup on all modal triggering buttons I could find
- Some other cleanup


app/assets/javascripts/gallery.js.erb
```diff
@@ -655,6 +655,7 @@ $(document).ready(function(){
   $('.delEnvironment').on('click', function(){
     var name = encodeURIComponent($(this).attr('data-environmentName'))
     $('#confirmationModal').modal('show');
+    makeScreenreaderAlert('modalAlert', 'Dialog has appeared.');
```
mcrutch, line 658: With the on(bs.dialog.show) action wouldn't all of these be duplicates/redundant?

manfromjupyter: I thought so and it does work for the vast majority but it wasn't working for these particular ones because of the `return false` at the end of each of those functions which I did not touch out of fear those falses were needed for other reasons... to prevent other functions and processes from being called or modifying the page further. Would you like me to dive into each of these to see if we can remove that `return false`?

mcrutch: probably just need an e.preventDefault(); rather than return false; so see if that resolves the issue.

manfromjupyter: Will do

manfromjupyter:
@mcrutch you were right... As always. That show vs shown did the trick. This is ready now.

PR #920, opened 2023-09-14
Allow for deletion of groups without any notebooks

### LINKED ISSUE #917: Allow for Group Deletion
Probably only empty groups.

mcrutch:
Closes #917

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -141,8 +141,8 @@ class GroupsController < ApplicationController
       @group.destroy
       flash[:success] = message
       head :no_content
-    else(@group.owners.include?(@user))
-      render json: { message: "Group deletion failed. The group still contains notebooks" }, status: :unprocessable_entity
+    else (@group.owners.include?(@user))
+      render json: { message: "Group deletion failed. The group still contains notebooks." }, status: :unprocessable_entity
     end
   end
 
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -136,9 +136,14 @@ class GroupsController < ApplicationController
 
   # DELETE /groups/:gid
   def destroy
-    @group.destroy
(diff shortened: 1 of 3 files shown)


app/views/modals/_manage_group.slim
```diff
@@ -67,4 +67,6 @@ div.modal.fade id="manageGroup" aria-labelledby="manageGroupHeader" aria-describ
           div.modal-footer
             div
               button.btn.btn-danger type="button" data-dismiss="modal" Cancel
+              -if @group.notebooks.count == 0
+                button.btn.btn-danger id='DeleteGroup' type="button" Delete Group
```
manfromjupyter, line 71: Are we okay with making the id="deleteGroup"? Double quotes and camel case of the ID with first letter being lowercase to match the rest? I know we'd need to change the JS line for it too.

mcrutch: yeah yeah

app/controllers/groups_controller.rb
```diff
@@ -136,9 +136,14 @@ def landing
 
   # DELETE /groups/:gid
   def destroy
-    @group.destroy
-    flash[:success] = "Group has been destroyed successfully."
-    head :no_content
+    if @group.notebooks.count == 0
+      message = "Group #{@group.name} has been deleted successfully."
+      @group.destroy
+      flash[:success] = message
```
manfromjupyter, line 142: I feel like we should add a JSON response also to this on success that uses the message but I know this will be covered eventually by the one ticket to refactor all of that and ensure our system is returning the messages both in the UI and outside of it.

manfromjupyter (approved):

app/controllers/groups_controller.rb
mcrutch: No point in doing a JSON return on success because you have to navigate away from the group's page since it doesn't exist anymore, so the flash message suffices as it navigates to the group listing again.

Issue #853, opened 2023-05-18
Replace ".hide" with hidden


manfromjupyter:
.hide class only hides things visually but not for blind and low visions people using assistive technologies (screen readers).

`hidden` ex. `<p class="class1" hidden>Test</p>` will hide it both visually and for screen readers.

Places where we are applying .hide should instead be replaced with `hidden` and JS instead of adding/removing said class should be apping removing `.prop('hidden',true')`

## While at it
Places were we are using `style="display;none"` and using JS to toggle this, should also be replaced with this property. Hidden has been supported since one of the first few versions of Chrome, Edge, and Firefox and is the preferred cleaner way to do this: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden

mcrutch:
Some of the display:none stuff is baked into bootstrap dropdowns and would take a lot of effort to override and undo.

PR #922, opened 2023-09-14
change from .hide to hidden="true"

mcrutch:
Relates to #853 
Only does the .hide part
the display:none is going to be a bit more involved.


manfromjupyter (approved):
Looks good and works great. Confirmed on JAWS and in a styleless view the app is flawless with these hiddens.

As per Bootstrap, because we can't change that I'd say let's ignore that. I'm okay with closing if you are.

PR #926, opened 2023-09-18
Fix a trusted tag edge case

### LINKED ISSUE #925: If an admin ads the trusted tag, and stops being an admin, the tags can't be updated by non-admins
This is because it checks if the trusted tag which exists on the notebook is owned by a current admin when trying to update tags.

mcrutch:
Use Case:
- User A is an admin
- User A adds Trusted to a notebook
- User A stops being an admin
- User B (admin or not) tries to add tag "foo" to a notebook
- User B would be blocked because the trusted tag is no longer valid

How it's fixed
- tag.invalid is only checked for NEW tag records added to the notebook

Closes #925


manfromjupyter (approved):

manfromjupyter:
Good catch!

Issue #850, opened 2023-05-16
Tag Add/Edit/Delete Refactor


manfromjupyter:
1. Support persons with disabilities
2. Fix issue where user adds text to tag field to add but submits form.

Ideally split up the tags into 2 things. Place to add and display of the tags (not in the same field).

mcrutch:
Relates to #415

mcrutch:
Does #935 resolve this in your mind @manfromjupyter ?

manfromjupyter:
@mcrutch sorry for the delay. It's definitely an improvement but I see the following things that should probably get addressed if it's even possible.

My quick summary:

## All Tag Fields
1. When a tag is deleted or added we need a screenreader alert makeScreenreaderAlert() or whatever saying as such. 
2. Within the label of that additional "Enter Tags here" field, we should mention to screenreaders, "Enter space key or enter key to add tags"
3. On the tags visual/deletion field, it should read to screenreaders in the label that Enter or ESC or whatever will delete tags. I know we added text to the x button but they can still accidentally delete with the backspace character or while reading the tags itself could press "ENTER" which would delete them.

## Tag Field just on Upload Modals
1. aria-label is "undefined" for some reason so it reads "Undefined edit Enter tags here type of text". 

## Tag Field just on Notebooks
1. No placeholder text

Issue #415, opened 2021-01-12
Need some tweaking on the tags tokenizer field of upload modal and maybe also search filtering (Accessibility)


manfromjupyter:
If you have the tags "marketing", "test", and "data", it will read to screenreaders as "marketingx testx datax" because the delete icon is displayed/written in code as just the letter "x". Need to make the "x" aria-hidden and add screenreader only text and replace with "remove tag". While at it should ideally changed from a link (<a> tag) to a button.

mcrutch:
Related #526

mcrutch:
I think we are done here. Concur @manfromjupyter ?

PR #935, opened 2023-09-25
Improve tokenization accessibility

mcrutch:
Also demo how to work around form submit without fully tokenizing
Relates to #415 
Relates to #850


manfromjupyter (approved):
Does not fully support BLV folks needing screen readers because pressing keys on the "Remove value" part of the tags, the X doesn't delete it BUT this is a massive and beautiful improvement. I recommend we keep #850 open or make a another one to capture this last remaining thing.

PR #948, opened 2023-10-20
Admin Ability to Edit/View/Delete User Environments

### LINKED ISSUE #942: Admin interface to view/edit user environments
Sometimes users screw up their environment URL, and right now for admins to even be able to see it they have to go into the database.  Make an interface so admins can see what environments a user has configured AND change them as well.

manfromjupyter:
Closes #942 

Also fixes a long time tech debt issue of environments modals not being within our modals section AND environment URLs using non absolute paths.


app/controllers/environments_controller.rb
```diff
@@ -45,20 +45,23 @@ def edit
 
   # POST /environments
   def create
+    user = check_user(params[:user_id].to_i, 'You are not allowed to create an environment for this user.')
```
mcrutch, line 48: You can't add any new required fields to this, params[:user_id] cannot be assumed to be present.

mcrutch: I guess if you want further clarification, it is because Jupyter Classic and Jupyter Lab extensions use this same exact endpoint

mcrutch:
I have at least confirmed so far that environments created by the Jupyter extensions still works with these changes, have more testing to do obviously.

mcrutch:
Only slightly awkward thing is that there are now two distinct URLs to get to my own environment listing.
domain.com/users/1-username/environments
vs 
domain.com/environments

But that might be a necessary evil now.

mcrutch:
Icon color oddity
image

manfromjupyter:
> image Handle this error better. Not a blocker can be put in another ticket if everything else checks out, but want to document.

environment messages cleaned up

manfromjupyter:
Rest is ready now

PR #949, opened 2023-10-23
Make import notebook form disable submit

mcrutch:
fix admin import notebooks form button disable
relates to #848


mcrutch:
Tried to muck with the login/registration modals, but really not worth while.  Also not sure what you want to do with Search exactly.

manfromjupyter (approved):

manfromjupyter:
> Tried to muck with the login/registration modals, but really not worth while. Also not sure what you want to do with Search exactly.

@mcrutch approved. As per the above comment, apologies. It's not search that need handling, it's the search filters. In my opinion the "Add Filter" button should be disabled by default and Search button should always be enabled BUT should only search with what filters have been added. After any time any filter is added, the button should then be disabled again.

PR #950, opened 2023-10-23
More tag accessibility tweaks

mcrutch:
Could be the last updates needed for #935 
- Announces add/remove tags
- Adds extra screen reader text (Need you to confirm the backspace part)
- Fixed two individual forms that were broken


app/views/application/_tags_edit.slim
```diff
@@ -2,7 +2,7 @@
 div.tags id="tagsEdit" style="display: none"
   form id="tagsEditForm" enctype="multipart/form-data"
     div.input-group data-toggle="tooltip"
-      input.form-control id="editTags" aria-label="Enter tags here" type="text" value=notebook.tags.map(&:tag_text).join(",") name="tags"
+      input.form-control id="editTags" aria-label="Enter tags here (Press space key or enter key to add tags)" placeholder="Enter tags here" type="text" value=notebook.tags.map(&:tag_text).join(",") name="tags"
```
manfromjupyter, line 5: Because it reads out the special characters as "left paran" and "right paran", would we be open to changing the aria-label to just:

```html
aria-label="Enter tags here. Add using space or enter keys."
```

manfromjupyter:
Can we also add this functionality to the admin import tags as well?

After that and the label recommended edit, approved.

PR #951, opened 2023-10-23
This was triggering on both keydown AND keyup for some reason, this w…

### LINKED ISSUE #936: Advanced Search invalid warning
- Go to search
- Click advanced search
- select a field
- select a filter
- enter text
- press `<tab>`
- press `<space>` (`<enter>` works, clicking the button also works)

Filter adds but displays error that Value is empty
image

mcrutch:
…ill make it only trigger on keydown.

Closes #936


app/assets/javascripts/gallery.js.erb
```diff
@@ -334,6 +334,7 @@ $(document).ready(function(){
     toggleAriaExpanded(element);
   });
   $('#filterFormSubmit').on('click', function(e){
+    console.log("FAIL");
```
manfromjupyter, line 337: do we still want this console.log?

manfromjupyter (commented):
@mcrutch Because of the `e.preventDefault()` I think on this method, tabbing no longer works on the site for keyboard-only Users. Only lets you tab 1 time. Is there another way we can fix the issue you're running into?

manfromjupyter (approved):

PR #960, opened 2023-11-03
Ruby 3.2

mcrutch:


mcrutch:
3.2 is going to be a heavy lift, this alone won't cut it.

PR #962, opened 2023-11-08
Update gems

mcrutch:
Trying to resolve some warnings during the asset pipeline build, this solves that but I had to make some changes to the asset pipeline and dragged a bunch of other gems along to newer versions as well.


blueninja416 (approved):

config/initializers/assets.rb
mcrutch, line 1: These are picked up by the other directives in manifest.js

app/assets/javascripts/cable.js
mcrutch, line 1: We aren't using cable, so the channels directory doesn't exist, so this killed asset precompile until commented out

app/assets/config/manifest.js
mcrutch, line 1: pre-compile the fonts this way rather than the other

PR #964, opened 2023-11-14
Deprecated Notebooks can't set Alt Notebook as itself

### LINKED ISSUE #958: You can set a notebook as its own alternate when deprecating a notebook
Probably shouldn't be able to specify the notebook itself or another deprecated notebook as the alternate for a deprecated notebook.

manfromjupyter:
Closes #958 

Not the perfect solution. Perfect solution would also exclude the notebook from appearing in recommendations at all and then cause front end errors now letting users lose the deprecation reasoning text they typed if that failed. That all being said, here is the quick solution to prevent saving deprecation status from both the UI and from console users when the alternate notebook ID contain the notebook's ID itself.


blueninja416 (approved):
This all looks good to me!

PR #971, opened 2023-11-21
Notebook Privacy Changes no longer Affects Review Status

### LINKED ISSUE #730: Verified notebooks lose verification (approved reviews) if you make public/private

manfromjupyter:
Closes #730


app/models/review.rb
```diff
@@ -40,7 +40,7 @@ def recommended_reviewer?(user)
 
   # Is this review "recent"?
   def recent?
-    latest_revision_id = notebook.revisions.order(id: :desc).first.id
+    latest_revision_id = notebook.revisions.order(id: :desc).where.not(revtype: "metadata").first.id
```
blueninja416, line 43: It looks like metadata isn't a supported `revtype` according to the `reviewable_by` function above, is that meant to be allowed here?

manfromjupyter: Shoot I think you're right. I think it was meant to be the "type" type lol

manfromjupyter: @blueninja416 I just looked into it and it looks like it's right after all. `revtype` is used on Reviews and is limited to the types of `reviewable_by` but the Revisions database table also uses it but for types `create`, `update`, and `metadata`.

Example of what database might look like if someone made a private notebook and then made it public:
```
+------+---------+-------------+--------+------------------------------------------+----------+---------------------+---------------------+----------------+-------------------+----------------+
| id   | user_id | notebook_id | public | commit_id                                | revtype  | created_at          | updated_at          | commit_message | change_request_id | friendly_label |
+------+---------+-------------+--------+------------------------------------------+----------+---------------------+---------------------+----------------+-------------------+----------------+
| 2179 |       1 |         538 |      0 | 729ebac56bdbd6f442aeca1708c9d0de5aed961d | create   | 2021-01-19 15:21:34 | 2021-01-19 15:21:34 | NULL           |              NULL | NULL           |
| 2181 |       1 |         538 |      1 | 729ebac56bdbd6f442aeca1708c9d0de5aed961d | metadata | 2021-01-21 12:46:28 | 2021-01-21 12:46:28 | NULL           |              NULL | NULL           |
+------+---------+-------------+--------+------------------------------------------+----------+---------------------+---------------------+----------------+-------------------+----------------+
```

PR #972, opened 2023-11-21
Carry Forward Queued Reviews Config

### LINKED ISSUE #578: Carry forward Queued Reviews on notebook updates

manfromjupyter:
Closes #578 

Queued reviews carry over to next versions if all reviews are queued AND the config to allow it is set to true. I recommend making it default functionality but adding it as a config just in case people want it.

### DIFF
--- a/app/controllers/notebooks_controller.rb
+++ b/app/controllers/notebooks_controller.rb
@@ -160,6 +160,7 @@ class NotebooksController < ApplicationController
       if summary.length > 250
         errors += "Change log was too long. Only accepts 250 characters and you submitted one that was #{summary.length} characters. "
       end
+      previous_revision = Revision.where(notebook_id: @notebook.id).last.id
     end
     if errors.length <= 0 && save_update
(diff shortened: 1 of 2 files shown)


blueninja416 (approved):
Everything makes sense to me!

PR #974, opened 2023-11-21
Reviews/Change Request Table consistency with Review/Change Request "All" pages

### LINKED ISSUE #426: Make user's reviews page and user's change requests page tale format look the same as the /reviews and /change_requests/all pages

manfromjupyter:
Closes #426 

Made Reviews Index and Change Request Index filtered by title the same way All Reviews and All Change Requests does

Have NOT confirmed yet that it works with custom title parsing/configs but out-of-the-box experience works and I am confident customs should work.

### DIFF
--- a/app/views/application/_reviews.slim
+++ b/app/views/application/_reviews.slim
@@ -30,16 +30,21 @@
         th Reviewer
         th Time
     tbody
-      -sorted_array = reviews.sort {|a,b| b.updated_at <=> a.updated_at }
+      -if GalleryConfig.notebook_title_parse.parsing_enabled
+        -sorted_array = reviews.order(:updated_at).sort {|a,b| a.notebook.title.partition(GalleryConfig.notebook_title_parse.after_character)[2].downcase <=> b.notebook.title.partition(GalleryConfig.notebook_title_parse.after_character)[2].downcase }
+      -else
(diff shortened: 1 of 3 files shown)


app/views/application/_reviews.slim
```diff
@@ -30,16 +30,21 @@
...
           -next
         -else
           tr
-            td
-              -if !review.notebook.public? && (@user.admin? || @user.can_edit?(review.notebook))
-                ==image_tag("Lock.png", class: "tagLogoLock tooltips", alt: "Private Notebook", title: "This notebook is private", tabindex: "0")
-              ==link_to_notebook(review.notebook)
+            td data-sort="#{render partial: GalleryConfig.slim.table_notebook_title_text_only, locals: { notebook: review.notebook }}"
+                ==render partial: GalleryConfig.slim.table_row_heading_label, locals: { notebook: review.notebook }
+                -if !review.notebook.public? && (@user.admin? || @user.can_edit?(review.notebook))
```
blueninja416, line 44: Is the `(@user.admin? || @user.can_edit?(review.notebook))` check needed here? Looking at line 38 it seems if they don't meet these conditions they won't see the notebook anyways.

blueninja416 (approved):
Looks like everything should be good, there's that extra check that I'm not sure is an issue or not, but other than that everything makes sense.

PR #1003, opened 2025-01-30
sets the correct sorting `title` for mysql

mild-bit:
If the user searches for something, the sorting order is done with Solr instead of MySQL. MySQL uses `title` and Solr uses `title_sort` so this needs to be specified.


ace6150 (approved):
Is there any way to unit test the title sort? That's my only comment.

eldenlord204824 (approved):

PR #1004, opened 2025-02-06
log invalid auth token events

mild-bit:
ExceptionNotifier plugin ignores InvalidAuthenticityToken exceptions. Also added more sections to the emails.


eldenlord204824 (approved):
Just leaving my thoughts lowside too. I still think it's a good idea to explore making a flag feature that blocks certain common exceptions from being emailed instead of out right saying, no you can't get these emails outright. This is a fine temporary solution though.

ace6150 (approved):

PR #976, opened 2023-11-22
Remove/Add Recommended Reviewers workflows for Reviews

manfromjupyter:
Closes #190

NOT DONE YET

## Proposed View as Notebook Editor (and admin)
Reviews page as notebook editor and admin

## Proposed View as Recommended Reviewer
Reviews page as recommended reviewer

PR #1007, opened 2025-03-10
Notebook Verification Changes

eldenlord204824:
## Main Changes:
- notebook entries in database are now explicitly verified or unverified in their table
  - there is also a script that can be run for DBs that already have existing verified notebooks to have them be labeled as such in the table
- users can filter the search and all notebooks to only show verified notebooks
- verified notebooks are directly boosted with an increased score when searching for notebooks.

## Other Comments:
- This is all the big logical stuff I could think of doing with verified notebooks - open to suggestion or additional ideas
- I am thinking that the homepage notebooks need to also be showing verified notebooks in the same way search does, but that can be a separate issue/ticket since this relies much more on manipulating the score based (which I need to think more on) on the recommended notebooks for a specific user.

### DIFF
--- a/script/verify_notebooks.rb
+++ b/script/verify_notebooks.rb
@@ -1,4 +1,4 @@
-# To run this script: rails runner script/create_user.rb
+# To run this script: rails runner script/verify_notebooks.rb
 
 puts "Verifying notebooks not properly verified..."
 
--- a/script/verify_notebooks.rb
+++ b/script/verify_notebooks.rb
@@ -3,7 +3,7 @@
 puts "Verifying notebooks not properly verified..."
 
 Notebook.find_each do |nb|
-  nb.set_verification(review_status == :full)
+  nb.set_verification(nb.review_status == :full)
   nb.save if nb.changed?
 end
 
--- a/script/verify_notebooks.rb
+++ b/script/verify_notebooks.rb
@@ -1,0 +1,12 @@
+# To run this script: rails runner script/create_user.rb
+
+puts "Verifying notebooks not properly verified..."
+
+Notebook.find_each do |nb|
+  if nb.review_status == :full
+    nb.toggle_verification
+    nb.save if nb.changed?
+  end
+end
+
+puts "Complete"
--- a/script/verify_notebooks.rb
+++ b/script/verify_notebooks.rb
@@ -3,10 +3,8 @@
 puts "Verifying notebooks not properly verified..."
 
 Notebook.find_each do |nb|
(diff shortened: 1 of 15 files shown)


ace6150 (approved):
Marking as approved, but I admit I don't have a great understanding of what is happening.

mild-bit (commented):
Probably should have mentioned this in [add-remove-rec-reviewers](https://github.com/nbgallery/nbgallery/pull/976) but are we not going to enable reviews _here_ ? I believe we should have them on by default as we move towards a more review focused app. I'm open to hearing your thoughts about this.

eldenlord204824:
> Probably should have mentioned this in [add-remove-rec-reviewers](https://github.com/nbgallery/nbgallery/pull/976) but are we not going to enable reviews _here_ ? I believe we should have them on by default as we move towards a more review focused app. I'm open to hearing your thoughts about this.

I agree, but not at the present. Once we are confident we have a structure that is satisfying, which I am not quite at that confidence level yet, we probably should enable by default.

mild-bit (changes requested):

mild-bit (commented):
When I click on `Show Only Verified` I still get the same output. is this the expected behavior?
Screenshot from 2025-03-21 13-16-50

Screenshot from 2025-03-21 13-18-57

script/verify_notebooks.rb
```diff
@@ -0,0 +1,10 @@
+# To run this script: rails runner script/verify_notebooks.rb
+
+puts "Verifying notebooks not properly verified..."
+
+Notebook.find_each do |nb|
+  nb.set_verification(review_status == :full)
```
mild-bit, line 6: nb.review_status == :full *

mild-bit (changes requested):

eldenlord204824:
No, the behavior in the screenshots is not supposed to happen. Mine is working fine. The way that displays the given filters is strictly though the database not SOLR. Is the database updated? Did you do it through the app or change the database directly?

Capture
Capture2

mild-bit (commented):
The database was updated after I ran the `verify_notebooks.rb` script. 
Screenshot from 2025-03-24 17-02-47

eldenlord204824:
I made some slight changes to the code (same filter functionality) and added a filter to the display of a certain group's notebooks. I am not sure what is going on with filtering and where the error is coming from or if the changes make a difference (likely not)...

mild-bit (approved):

Universe-sudojam (approved):

PR #1020, opened 2025-05-01
Adding new restrictions for notebook reviews

eldenlord204824:
Reviews can no longer be proposed for notebooks in a deprecated state and private notebooks are now optional to block from review proposals.


eldenlord204824:
After thinking about it a bit, should we make the propose button disappear? Or should it just be a banner? It may be more user friendly to just have it be an error banner... Thoughts?

Universe-sudojam:
> After thinking about it a bit, should we make the propose button disappear? Or should it just be an error banner? It may be more user friendly to just have it be an error banner... Thoughts?

I think an error banner would be better for user awareness

mild-bit (commented):
What should happen in the event when a public notebook is up for review and the notebook owner changes the notebook to private when `enable_private_notebook_reviews: false`?

mild-bit:
> > After thinking about it a bit, should we make the propose button disappear? Or should it just be an error banner? It may be more user friendly to just have it be an error banner... Thoughts?
> 
> I think an error banner would be better for user awareness

I'm under the opinion that the `Propose for Review` button should just disappear for deprecated notebooks. If its a notebook function that the user cannot use then it should not appear in the notebook actions dropdown.

eldenlord204824:
> What should happen in the event when a public notebook is up for review and the notebook owner changes the notebook to private when `enable_private_notebook_reviews: false`?

any reviews that don't directly impact the status of a notebook (approved or unapproved) are deleted completely.

eldenlord204824:
> > > After thinking about it a bit, should we make the propose button disappear? Or should it just be an error banner? It may be more user friendly to just have it be an error banner... Thoughts?
> > 
> > 
> > I think an error banner would be better for user awareness
> 
> I'm under the opinion that the `Propose for Review` button should just disappear for deprecated notebooks. If its a notebook function that the user cannot use then it should not appear in the notebook actions dropdown.

I am more inclined to agree with the fact the button should just disappear since it isn't an available option for deprecated (and possibly private) notebooks. It would just be up to communicate that to users.

mild-bit (approved):

PR #1022, opened 2025-05-08
Hiding unapproved notebooks from public view

eldenlord204824:
1. Unapproved notebooks should not be viewable unless filtered.
2. New boolean field for notebooks.

### DIFF
--- a/app/assets/javascripts/gallery.js.erb
+++ b/app/assets/javascripts/gallery.js.erb
@@ -397,17 +397,46 @@ $(document).ready(function(){
     $('#filterVerifiedNotebooksForm #onlyVerifiedNotebooksCheckbox').click();
   });
 
+  $('#showUnapprovedToken, .tokenfield #filterUnapprovedForm').click(function(){
+    $('#filterUnapprovedForm #unapprovedCheckbox').click();
+  });
+
   $('#showPrivateToken, .tokenfield #showPrivateNotebooksCheckbox').click(function(){
     $('#filterPrivateNotebooksForm #showPrivateNotebooksCheckbox').click();
   });
 
   // Filter search results by those that are/aren't deprecated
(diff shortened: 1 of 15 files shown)


mild-bit (commented):
I'm still able to see unapproved notebooks whether I click on the `Show unapproved` checkbox or not.
Screenshot from 2025-05-09 15-02-12

eldenlord204824:
> I'm still able to see unapproved notebooks whether I click on the `Show unapproved` checkbox or not. Screenshot from 2025-05-09 15-02-12

Mine works did you reindex? did you unapprove those before checkout out th branch?

Unapproved hidden:
unapproved hidden

Unapproved showing:
unapproved shown

mild-bit (commented):
Another note: These checkboxes in particular are very buggy. Whenever I click on the checkbox it either does nothing after the page refreshes or actually toggles as intended.

Screenshot from 2025-05-09 15-41-48

mild-bit (commented):
Question: What would or should happen when a user checks on `only verified` and `show unapproved`? Don't these contradict each other?

eldenlord204824:
> Question: What would or should happen when a user checks on `only verified` and `show unapproved`? Don't these contradict each other?

show only verified. both deprecated and unapproved notebooks can't be either of those statuses and verified. Right now the backend is (assuming all checkboxes are marked) querying notebooks that are verified and unapproved and deprecated, which is impossible so only verified are shown. It may be beneficial to unmark unapproved and/or deprecated when verified is marked.

Other thing I just thought of as well. I am not sure if unapproved notebooks should be "deprecatable". they essentially mean the same thing just one is based on reviewer criteria and the other based on owner criteria...

mild-bit (approved):
Ok, it looks like I had to reindex. Everything looks good.

PR #1064, opened 2026-07-07
fix: wrap db transactions to writer role

mild-bit:
Fix: Wrap any db transactions to the `writer` role. 

Example: 
```ruby
ActiveRecord::Base.connected_to(role: :writing) do
  NotebookSimilarity.transaction do
    NotebookSimilarity.where(notebook_id: notebook.id).delete_all # no callbacks
      NotebookSimilarity.import(records, validate: false)
  end
end
```

### DIFF
--- a/app/controllers/user_preferences_controller.rb
+++ b/app/controllers/user_preferences_controller.rb
@@ -2,7 +2,9 @@ class UserPreferencesController < ApplicationController
   before_action :verify_login
 
   def index
-    @user_preference = UserPreference.find_or_create_by(user_id: @user.id)
+    ActiveRecord::Base.connected_to(role: :writing) do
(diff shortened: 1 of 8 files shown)


eldenlord204824 (approved):

eldenlord204824 (commented):
There is more erroring out after digging. Might be good to dig and look for any transactions that modify DB contents.

Image

eldenlord204824 (changes requested):

mild-bit:
> There is more erroring out after digging. Might be good to dig and look for any transactions that modify DB contents.
> 
> Image

ok, i'll take a closer look at this

eldenlord204824 (approved):

